PR #2, opened 2016-01-04
Feature #10864 Enhance installation section,

helge000:
--
DANIEL HELGENBERGER <daniel.helgenberger@m-box.de>

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,12 @@ Requirements
 Installation
 ============
 
+Clone this repository in your icingaweb2 `modules` directory and name it `director`, eg.:
+
+```
+git clone https://github.com/Icinga/icingaweb2-module-director.git /usr/share/icingaweb2/modules/director
+```
+
 Create Icinga Director database
 -------------------------------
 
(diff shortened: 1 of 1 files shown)


tux-00:
1. Since we create the database with `CHARACTER SET 'utf8'` maybe we need to set `charset = "utf8"` in the resources.ini ?

2. Maybe we can tell to import the `schema/mysql.sql` file in mysql with `mysql -u root -p director < mysql.sql`

3. Review the MySQL part of the README.md with more steps like (create db, import mysql.sql file, grant access) ?

Thomas-Gelf:
Hi,

@helge000: sorry for not answering earlier. I had other plans with the setup and therefore documentation in my local stash diverged from master for a long time.

@tux-00: charset = "utf8" -> that's correct, Director currently complains if this is not set on it's module config page and will completely refuse to run in future. Encoding matters. Documentation shows this in the sample ini file and contains a note explicitly mentioning this.

The schema should NOT be manually imported, please check documentation in the current master. Especially the "automation" part. Director handles database migrations on it's own, users should not be taught to manually mess with SQL files. You can, of course - that's why we ship them. But just running:

`icingacli director migration run`

...is definitively less painful and error-prone. You could basically run this once a minute via cron, twice an hour via Puppet or just on demand. It will create a missing schema and upgrade a changed one.

Director will also VERY soon offer to trigger migrations through the web frontend.

Please let me know if there are still related doubts.

Thanks,
Thomas

tux-00:
Ok for the encoding.

When I'm trying to follow the README.md file, I type `mysql -e "CREATE DATABASE director CHARACTER SET 'utf8'; GRANT ALL ON director.* TO director@localhost IDENTIFIED BY 'some-password';"` and the following error appears : `ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)`

So I'm connecting to mysql with `mysql -u root -p`, I type : 
```
mysql> CREATE DATABASE director CHARACTER SET 'utf8';
Query OK, 1 row affected (0.07 sec)

mysql> GRANT ALL ON director.* TO director@localhost IDENTIFIED BY 'some-password';
Query OK, 0 rows affected (0.01 sec)
```

Is the GRANT command not working because of no table in director db ?

Now when I'm going to my Icingaweb I get this error : 
```
Zend_Db_Statement_Exception in /usr/share/php/Zend/Db/Statement/Pdo.php:235 with message: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'director.icinga_host' doesn't exist
```

I don't know how the database configuration of the module is working.

Thomas-Gelf:
You are using -p telling MySQL to ask you for the password when connecting to the DB, but there is no "-p" in your "-e" call. Thats why the error message also reports "using password: NO". Once the db exists please run

`icingacli director migration run --verbose`

as shown here:

https://github.com/Icinga/icingaweb2-module-director/blob/master/doc/30-Automation.md

And your errors will go away. As mentioned before, this will also be executable via the GUI very soon, so instead of the exception you see right now it will ask very friendly "Should I create the database schema for you?".

In the meantime you have to follow the instructions linked documentation for manual/automated installations. 10 seconds after you launched schema migration helper, the exceptions will vanish from your web frontend, no reload required.

Cheers,
Thomas

helge000:
> sorry for not answering earlier

no worries! Better close it then!

PR #6, opened 2016-03-09
Update Db.php

vbalkanov:
Fix error in public function fetchActivityLogEntry($checksum)


Thomas-Gelf:
Oops... committed, thanks. Might take another 3 minutes unless it gets replicated to GitHub.

Thanks,
Thomas

PR #8, opened 2016-03-31
Update FormExtensibleSet.php

vbalkanov:
Change Add, Remove, Up & Down icons from hieroglyph to normal.
Before:
http://s017.radikal.ru/i444/1603/f8/df261c500ac9.jpg
After patch:
http://i053.radikal.ru/1603/a4/8455ed67db1c.jpg


Thomas-Gelf:
There is something going wrong with your styles, I guess you might be affected by https://dev.icinga.org/issues/11465. Did you try to SHIFT-Reload in your browser? It should work fine even without your patch. Another explanation could be that you're running Icinga Web 2 < 2.2.0. Director looks terrible on your screenshots ;)

Cheers,
Thomas

vbalkanov:
Oooops :)

PR #12, opened 2016-06-04
Housekeeping: Adapt DELETE commands for Postgresql

rzerres:
Hey Team,

I do use icingaweb2-module-director on a Postgresql Instance (v9.5.2). While testing, a master-satellite-node constallation i have generated lots of db-entries. Wanted to clean up a bit, withouth starting from scratch. So I got around the housekeeping functions. 

Hence until latest commit  in master (85cda6da9bbd78af6fc893aaf1d93c09a988db29) all functions that actually would delete rows (eg 'icingacli help housekeeping run unusedFiles') will return with an error. The counting functions work out as expected (Count column is filled correct). 

Please review my pull request. It will fix the job for my installation.
Postgres does not directly support an INNER JOIN on Delete requests. I pached the code in the following manner: When functions are called, check the connection Database Type (getDbType) and communicate with adequate syntax.  

Any comments are welcome.
Ralf


Thomas-Gelf:
Sorry, I completely forgot about this as there was no related issue on https://dev.icinga.org. I remembered that "someone told me" about such problems, but I wasn't able to find any trace of it. Today someone opened related PostgreSQL issues and I got a notification for a completely unrelated pull request from Github - and so I finally stumbled over this again.

Thanks for reporting it and thank you for your patches, I'll review them ASAP.

Cheers,
Thomas

Thomas-Gelf:
Looks good to me at first sight. I have to re-read PostgreSQL docs, maybe you could also help me: why is the addition of WHERE imported_row.checksum = r.checksum required? I can remember that PostgreSQL isn't able to do this without that USING-thing, but this looks so ugly to me, so redundant. No way to shorten it? You don't have to, but I'd love to ;-)

rzerres:
I do agree, ugly and looks redunant. This USING clause is the way postgres overcomes the non working 'commenly uses' SQL clause to delete objects. I have got the code sample from digging postgres forums discusses as a common approach. You create a JOIN on the table itself (therfore the USING). Now you can address the WHERE with different names consulting the same table from a reference. Hence you mimic the INNER JOIN. 
Maybe there is more clean syntax using psql DELETE options. Sorry I havn't investigated it so far, since i wanted to get a proposel with working code in first step. Maybe it would be wise to ask someone with deeper postgres experience in the forum?

Cheers,
Ralf

PS: By the way, is it more appropriate to communicate on dev.icinga.org for any pull request?

Tuxdiver:
I tried the pull request, but it fails on one task:
`root@v-monitor (monitor):/usr/share/icingaweb2/modules/director# icingacli director  housekeeping  run ALL
ERROR: Zend_Db_Adapter_Exception in /usr/share/icingaweb2/library/vendor/Zend/Db/Adapter/Pdo/Abstract.php:271 with message: SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation "import_rowset_row" does not exist
LINE 1: ...OM imported_row USING imported_row AS r LEFT JOIN import_row...
                                                             ^`

the table import_rowset_row does not exist. Changing that to "imported_rowset_row" gives the next error:
`root@v-monitor (monitor):/usr/share/icingaweb2/modules/director#  icingacli director  housekeeping  run unlinkedImportedRows
ERROR: Zend_Db_Adapter_Exception in /usr/share/icingaweb2/library/vendor/Zend/Db/Adapter/Pdo/Abstract.php:271 with message: SQLSTATE[42703]: Undefined column: 7 ERROR:  column rsr.rsr_checksum does not exist
LINE 1: ...ksum WHERE imported_row.checksum = r.checksum AND rsr.rsr_ch...
                                                             ^`

I think, it should be "rsr.row_checksum" - but I am not sure about that.

Could you have a look at that?

Best regards,
Dirk.

rzerres:
your are right. Just another typo while I did copy the old SQL. Will check in the update.

Ralf

library/Director/Db/Housekeeping.php
```diff
@@ -201,7 +201,7 @@ public function wipeUnlinkedImportedRows()
                      . ' LEFT JOIN import_rowset_row AS rsr ON'
```
Tuxdiver: I think this should be "imported_rowset_row" ?!

rzerres: Damn it .... of course you are right! 
Pushed the typo correction....

rzerres:
Hey Tom,

i read that you are going to release 1.1.0 tomorrow. Any chance to see the PostgreSQL pull-request  inside?
Ralf

Thomas-Gelf:
This should read "today" :p Issues related to this pull request are still on the roadmap:

https://dev.icinga.org/versions/301

I'm doing my best to get it in, just struggling with a nasty little bug with higher priority right now.

Cheers,
Thomas

PR #13, opened 2016-06-17
Fatal error: Call to undefined method Icinga\Module\Director\Objects\…

Paul-J-Barrett:
…IcingaHost::importedObjects() in /usr/share/icingaweb2/modules/director/library/Director/Web/Form/DirectorObjectForm.php on line 76

### DIFF
--- a/library/Director/Web/Form/DirectorObjectForm.php
+++ b/library/Director/Web/Form/DirectorObjectForm.php
@@ -72,8 +72,8 @@ abstract class DirectorObjectForm extends QuickForm
                 $object->imports = $el->getValue();
             }
         }
-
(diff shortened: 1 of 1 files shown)


Paul-J-Barrett:
After deleting director database and recreating database and then doing fresh git pull of director started getting this error. No where do we see importedObjects defined.

Thomas-Gelf:
Thanks for reporting this and sorry for not merging your pull request, merely because of style and comments. Fixed it in current master by removing the related line. This got broken today as related logic has been externalized. This line formerly HAD a purpose, it triggered a refresh of an internal resolve cache.

PR #14, opened 2016-06-23
Feature 11332 support dependencies

mdetrano:
This branch is my implementation Icinga Dependency support in the Director module.

It is _close_ to ready for merge, if it looks ok, though it still needs:

Updated scripts for modifying the DB in Postgres (everything here is only for mysql/mariadb).
An icon graphic for the Index page.
The schema version (98) should be verified.

Should also check over changes I made for plurals handling (since it's Dependencies...not Dependencys) in the web controllers and rendering code, and some tweaks I made for rendering unresolved related services, which need to be looked-up with array-type keys.


Thomas-Gelf:
This looks very good to me, thank you for this pull request. Please forgive me that I'm unable to merge it into 1.1.0, as it will be released tomorrow. However, I'll try to bring this to master immediately afterwards.

Things that need to be addressed:
- Form will load slow with thousands of hosts and/or services I guess. I'll reflect about how this could be improved.
- I do not understand the change in DbObject.php, you are checking for a variable that cannot exist
- You have been forced to create some hacks to work around a few nasty things, I'll check what can be simplified and where some base libraries could be improved
- Depending on your screen resolution there is not enough space for host/service tabs
- We need to create related PostgreSQL schema migrations

Minor formatting issues should be cleaned up:
- especially superfluous newlines and misaligned property lists
- I prefer ' over " unless " is really necessary
- space after comma
- space before and after dots when doing string concatenation

Please try to validate your code with a recent version of PHP CodeSniffer like follows:
`/usr/bin/phpcs --standard=PSR2 --extensions=php --encoding=utf-8 application/ library/Director/ --report=full`

Thanks a lot
Thomas

mdetrano:
On the DbObject.php changes,

I ran into this issue, if I recall, specifically when doing the Render step to write out the Dependencies to files.  I believe this may have been an artefact of the other changes I made to help with related-object resolution (which in turns helps the API find related services).  PHP didn't like the attempt to access the variable which was not set in that case, and threw an error.  The change suggested got me passed the error, but, of course, there may yet be a better way of handling the problem.

TheFlyingCorpse:
:+1:  on this, it would help a lot to get this in Director :)

ktella:
I definitely appreciate all your efforts on development of director.  Please get this into the master branch.
Thank you.

Thomas-Gelf:
Hi @mdetrano,

wheew, seems that you got a little bit in trouble when merging/rebasing this :D To make things easier for you and for me in future please always rebase origin/master, when a merge happens you have to get rid of the merge commit via reset or similar. I had no idea how to clean this up, so I pushed most of this with slight changes (mostly trailing whitespaces, formatting) to a new feature branch [feature/support-dependencies-11332](https://github.com/Icinga/icingaweb2-module-director/tree/feature/dependencies-11332).

The only important thing I didn't push is resolveUnresolvedRelatedProperty, we need to find a better solution there. I also renamed the SQL migration script, this way you need to apply it manually in the meantime,  but it has the advantage of not causing you any merge conflicts when I push the next migration to master. Do you need any help with the PostgreSQL migration?

I didn't really test it right now, looks already pretty good at first sight. Some things that came to my mind:
* we should remove the "Dependencies" tab and only show templates I guess, similar to services
* tabs for single object dependencies might disappear on smaller resolution, there are too many right now

However, I have to play more with it. Didn't immediately understand some things like how to choose the parent service, or how to define something like "make all/a specific agent-based check(s) depend on the check for the agent itself", as apart from host-dependencies that's my personal most important use-case for dependencies.

If you manage it to clean up your branch that's also fine with me, there is no need for your pull request to be a single commit. However, it must be straight forward, with no merges in it. Regardless of whether you choose to freshly fork from the branch I pushed or to clean up yours, please make a manual diff of my branch with your one to figure out whether a) I forgot something and b) what kind of things I adjusted.

Thanks a lot for the good work!

Cheers,
Thomas

mdetrano:
Hello..


On 09/13/2016 12:23 AM, Thomas Gelf wrote:
>
> Hi @mdetrano <https://github.com/mdetrano>,
>
> wheew, seems that you got a little bit in trouble when 
> merging/rebasing this :D To make things easier for you and for me in 
> future please always rebase origin/master, when a merge happens you have
>

Er, yeah.  I got into something whacky and not being fully versed in git 
I was bouncing around for a while trying to get a local copy of that 
branch online again.  I believe at some point I pushed to "master" on my 
fork when I meant to push to that branch, and that was it.  I was hoping 
the previous pull request would remain unaffected...but I guess not.

At this point, should I attempt to sync my fork, then rebase the branch ?

(like, checkout master, merge {your remote}/master , checkout {feature 
branch} , rebase master ... )





> to get rid of the merge commit via reset or similar. I had no idea how 
> to clean this up, so I pushed most of this with slight changes (mostly 
> trailing whitespaces, formatting) to a new feature branch 
> feature/support-dependencies-11332 
> <https://github.com/Icinga/icingaweb2-module-director/tree/feature/dependencies-11332>.
>
> The only important thing I didn't push is 
> resolveUnresolvedRelatedProperty, we need to find a better solution 
> there. I also renamed the SQL migration script, this way you need to 
> apply it manually in the meantime, but it has the advantage of not 
> causing you any merge conflicts when I push the next migration to 
> master. Do you need any help with the PostgreSQL migration?
>
> I didn't really test it right now, looks already pretty good at first 
> sight. Some things that came to my mind:
>
>   * we should remove the "Dependencies" tab and only show templates I
>     guess, similar to services
>   * tabs for single object dependencies might disappear on smaller
>     resolution, there are too many right now
>

Yeah, whatever you think is necessary and clean for the interface. As 
for related properties...I had not even wanted to touch that area but 
when testing the REST or command line API, I hit those issues, so I just 
worked on it enough to get through that glitch.



--Marc

Thomas-Gelf:
Hi Marc,

> at this point, should I attempt to sync my fork, then rebase the branch ?

Your master looks clean to me. Rebase for your branch will however not work, as from now on it will always rebase all your merge commits. Will get worse, not better. I'd suggest to go one of the following ways:

a) create a new branch from master
* keep your branch as is for reference
* cherry-pick single commits from the current branch
* manually re-add what's not possible to pick (git diff one-branch another-one is your friend)

b) create a new branch from the branch feature/dependencies-11332 I created for you
* keep your branch as is for reference
* git diff both branches to figure out what changes I applied there (shamelessly in your name) and whether/what is still missing

Doesn't matter which solution you prefer, in any case please always start with a clean new branch from master and keep the existing one for reference. That way you can operate without the risk to loose anything of the good work ;-)

> As for related properties...I had not even wanted to touch that area but when testing the REST or command line API, I hit those issues, so I just worked on it enough to get through that glitch.

That helped me to realize that it's pretty tricky and cumbersome to work with them. I'll try to push a few related helper functions to master to make things easier.

Cheers,
Thomas

Lali1006:
When will Icinga2 Director support dependency configuration feature? Which version?

Lali1006:
When will Icinga2 Director support dependency configuration feature? Which version?

PR #17, opened 2016-07-25
Add missing permission objects

zearan:
https://dev.icinga.org/issues/11843

### DIFF
--- a/configuration.php
+++ b/configuration.php
@@ -1,10 +1,26 @@
 <?php
 
 $this->providePermission('director/api', $this->translate('Allow to access the director API'));
-$this->providePermission('director/hosts/read', $this->translate('Allow to configure hosts'));
+$this->providePermission('director/ApiUsers/read', $this->translate('Allow to see api user details'));
+$this->providePermission('director/ApiUsers/write', $this->translate('Allow to configure api users'));
+$this->providePermission('director/commands/read', $this->translate('Allow to see command details'));
+$this->providePermission('director/commands/write', $this->translate('Allow to configure commands'));
+$this->providePermission('director/endpoints/read', $this->translate('Allow to see endpoint details'));
+$this->providePermission('director/endpoints/write', $this->translate('Allow to configure endpoints'));
+$this->providePermission('director/hosts/read', $this->translate('Allow to see host details'));
 $this->providePermission('director/hosts/write', $this->translate('Allow to configure hosts'));
+$this->providePermission('director/notifications/read', $this->translate('Allow to see notification details'));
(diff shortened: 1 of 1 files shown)


Thomas-Gelf:
Hi zearan,

don't worry, issues on dev.icinga.org will not be forgotten :-) The current permissions have been implemented long time ago for a couple of quick tests, but they where not meant to be used like this in the long run. I'm currently working on v1.1.1, and there will be v1.2.0 in autumn. Not much time left, as there are also other tasks - but if things go as expected the permission system should finally be enforced with 1.2.0.

It will be very granular but not the way you addressed it in your patch. There should be a strong distinction between a "Director Admin" role and different level of users, without the need for one to distinct between which permissions are critical and which are not. So, templates, field assignment and commands will always be admin-only tasks, as they would always allow you to work around any restriction in some "creative" way.

Granular control would then mostly happen on a template level, and also based on upcoming host and service template sets. So, as far as it works in my head right now, you will grant specific user (groups) the permission to use only specific template (sets) with specific fields. It should allow you to say "my users must choose at lease one OS- and one Location-related Host-template. They are further allowed to choose amongst 20 different predefined Service-Check-sets and to tweak related thresholds and notifications".

I guess I understand why you wrote this patch, I guess you want to grant users access to Director without giving them "everything" in Icinga Web 2, is that assumption correct? If so, could you please let me know in more detail what you want to achieve? I do not want to implement all those single permissions as then I would probably not be able to get rid of them for a long time. But if there is anything I could do to help you with your current requirements while still fitting my vision I would try to prioritize those changes. At least if they would not involve too much work.

Cheers,
Thomas

zearan:
Hi Thomas,
Thank you for your reply. As you have guessed I simply wanted to prevent giving "Allow everything" in Icinga Web 2 to the accounts which we use for our day-to-day work and was not able to do so because the permission objects in my patch are already checked by Icinga Director but were not available to be granted in the Roles tab.
I am glad to hear that a full-featured permission system is about to be implemented as this is what I am looking for in the long run. I will try to list some of my core requirements below but from what I have read in your reply it looks like your vision is very close to this:

- Restrict all critical features to some sort of "admin" role
- Restrict roles to create new objects only from a set of templates (or their children)
- Restrict read/write and even viewing (!) access to existing hosts based on groups and/or templates (some sort of multitenancy)
- Make sure that apply rules cannot be matched to objects with no access. Maybe apply rules have to be a critical admin-only feature, I am not sure (Example: Notifications should not be sent to someone who has no access to the object)

Thank you for your hard work. I happily await every new feature in Icinga Director, so that one day I get rid of the leftover configuration files on the system. =)

PR #19, opened 2016-08-04
Allow override of blacklisted properties

ardeois:
It is now a separate class attribute which can be overridden.


Thomas-Gelf:
Merged to master, thanks!

PR #20, opened 2016-08-09
Add support for expressions

ardeois:
Any rendered string can contain variables encapsulated with "$$" characters

Example:
Display Name declared with `Port $$host.vars.tcp_port$$ check` will
be processed as `"Port " + host.vars.tcp_port + " check"`

API:
```bash
 ./director-curl POST director/service?name=my_service '{"display_name": "Port $$host.vars.tcp_port$$ check" }'
 ```

 Rendered config:
 ```
  apply Service "my_service" {
      import "my_template"

      display_name = "Port " + host.vars.tcp_port + " check"
  }
 ```

This feature will be needed to add support of Service Apply Rules. See https://dev.icinga.org/issues/11976

Custom handling with $$var$$ was prefered over native Icinga 2 macro support (`$var$` or `macro($var$)`) because macro doesn't resolve variables declared in apply for rules. 

E.g:
With `apply Service "name" for (config => host.vars.configs)`, `config` won't be resolvable with either `"$config$"` or `macro("$config$")`

refs #11976

### DIFF
--- a/test/php/library/Director/Objects/IcingaServiceTest.php
+++ b/test/php/library/Director/Objects/IcingaServiceTest.php
@@ -242,13 +242,13 @@ class IcingaServiceTest extends BaseTestCase
 
         $db = $this->getDb();
 
-        $service = $this->service('___TEST___service_$$not_replaced$$');
+        $service = $this->service('___TEST___service_$not_replaced$');
         $service->object_type = 'apply';
-        $service->display_name = 'Service: $$replaced$$';
+        $service->display_name = 'Service: $host.vars.replaced$';
         $service->assignments = array(
             'host.address="127.*"',
         );
-        $service->{'vars.custom_var'} = '$$replaced$$';
+        $service->{'vars.custom_var'} = '$host.vars.replaced$';
         $service->store($db);
 
         $service = IcingaService::loadWithAutoIncId($service->id, $db);
@@ -266,10 +266,10 @@ class IcingaServiceTest extends BaseTestCase
 
         $db = $this->getDb();
 
-        $service = $this->service('___TEST___service_$$not_replaced$$');
+        $service = $this->service('___TEST___service_$not_replaced$');
         $service->object_type = 'object';
-        $service->display_name = 'Service: $$not_replaced$$';
-        $service->{'vars.custom_var'} = '$$not_replaced$$';
+        $service->display_name = 'Service: $host.vars.not_replaced$';
+        $service->{'vars.custom_var'} = '$host.vars.not_replaced$';
         $service->store($db);
 
         $service = IcingaService::loadWithAutoIncId($service->id, $db);
--- a/test/php/library/Director/Objects/IcingaServiceTest.php
+++ b/test/php/library/Director/Objects/IcingaServiceTest.php
@@ -12,6 +12,7 @@ class IcingaServiceTest extends BaseTestCase
     protected $testHostName = '___TEST___host';
 
     protected $testServiceName = '___TEST___service';
+    protected $createdServices = [];
 
     public function testUnstoredHostCanBeLazySet()
     {
@@ -233,6 +234,51 @@ class IcingaServiceTest extends BaseTestCase
         );
     }
 
+    public function testVariablesInPropertiesAndCustomVariables()
+    {
+        if ($this->skipForMissingDb()) {
+            return;
+        }
+
+        $db = $this->getDb();
+
+        $service = $this->service('___TEST___service_$$not_replaced$$');
+        $service->object_type = 'apply';
+        $service->display_name = 'Service: $$replaced$$';
+        $service->assignments = array(
+            'host.address="127.*"',
+        );
+        $service->{'vars.custom_var'} = '$$replaced$$';
+        $service->store($db);
+
+        $service = IcingaService::loadWithAutoIncId($service->id, $db);
+        $this->assertEquals(
+            $this->loadRendered('service3'),
+            (string) $service
+        );
+    }
+
+    public function testVariablesAreNotReplacedForNonApplyObjects()
+    {
+        if ($this->skipForMissingDb()) {
+            return;
+        }
+
+        $db = $this->getDb();
+
+        $service = $this->service('___TEST___service_$$not_replaced$$');
+        $service->object_type = 'object';
+        $service->display_name = 'Service: $$not_replaced$$';
+        $service->{'vars.custom_var'} = '$$not_replaced$$';
+        $service->store($db);
+
+        $service = IcingaService::loadWithAutoIncId($service->id, $db);
+        $this->assertEquals(
+            $this->loadRendered('service4'),
+            (string) $service
+        );
+    }
+
     protected function host()
     {
         return IcingaHost::create(array(
@@ -242,10 +288,14 @@ class IcingaServiceTest extends BaseTestCase
         ));
     }
 
-    protected function service()
+    protected function service($objectName = null)
     {
+        if ($objectName === null) {
+            $objectName = $this->testServiceName;
+        }
+        $this->createdServices[] = $objectName;
         return IcingaService::create(array(
-            'object_name'  => $this->testServiceName,
+            'object_name'  => $objectName,
             'object_type'  => 'object',
             'display_name' => 'Whatever service',
             'vars'         => array(
@@ -279,7 +329,7 @@ class IcingaServiceTest extends BaseTestCase
                 }
             }
 
-            $kill = array($this->testServiceName);
+            $kill = $this->createdServices;
             foreach ($kill as $name) {
                 if (IcingaService::exists(array($name), $db)) {
                     IcingaService::load($name, $db)->delete();
--- a/library/Director/CustomVariable/CustomVariable.php
+++ b/library/Director/CustomVariable/CustomVariable.php
@@ -88,12 +88,12 @@ abstract class CustomVariable implements IcingaConfigRenderer
         return $this->modified;
     }
 
-    public function toConfigStringPrefetchable()
+    public function toConfigStringPrefetchable($renderExpressions = false)
     {
         if (PrefetchCache::shouldBeUsed()) {
-            return PrefetchCache::instance()->renderVar($this);
+            return PrefetchCache::instance()->renderVar($this, $renderExpressions);
         } else {
-            return $this->toConfigString();
+            return $this->toConfigString($renderExpressions);
         }
     }
 
--- a/library/Director/CustomVariable/CustomVariableString.php
+++ b/library/Director/CustomVariable/CustomVariableString.php
@@ -36,8 +36,8 @@ class CustomVariableString extends CustomVariable
         return $this;
     }
 
-    public function toConfigString()
+    public function toConfigString($renderExpressions = false)
     {
-        return c::renderString($this->getValue());
+        return c::renderString($this->getValue(), $renderExpressions);
     }
 }
--- a/library/Director/CustomVariable/CustomVariables.php
+++ b/library/Director/CustomVariable/CustomVariables.php
@@ -268,24 +268,24 @@ class CustomVariables implements Iterator, Countable, IcingaConfigRenderer
         return $this->storedVars;
     }
 
-    public function toConfigString()
+    public function toConfigString($renderExpressions = false)
     {
         $out = '';
 
         ksort($this->vars);
         foreach ($this->vars as $key => $var) {
             // TODO: ctype_alnum + underscore?
-            $out .= $this->renderSingleVar($key, $var);
+            $out .= $this->renderSingleVar($key, $var, $renderExpressions);
         }
 
         return $out;
     }
 
-    protected function renderSingleVar($key, $var)
+    protected function renderSingleVar($key, $var, $renderExpressions = false)
     {
         return c::renderKeyValue(
             $this->renderKeyName($key),
-            $var->toConfigStringPrefetchable()
+            $var->toConfigStringPrefetchable($renderExpressions)
         );
     }
 
--- a/library/Director/Db/Cache/PrefetchCache.php
+++ b/library/Director/Db/Cache/PrefetchCache.php
@@ -84,14 +84,14 @@ class PrefetchCache
     }
     */
 
-    public function renderVar(CustomVariable $var)
+    public function renderVar(CustomVariable $var, $renderExpressions = false)
     {
         $checksum = $var->getChecksum();
         if (null === $checksum) {
-            return $var->toConfigString();
+            return $var->toConfigString($renderExpressions);
         } else {
             if (! array_key_exists($checksum, $this->renderedVars)) {
-                $this->renderedVars[$checksum] = $var->toConfigString();
+                $this->renderedVars[$checksum] = $var->toConfigString($renderExpressions);
             }
 
             return $this->renderedVars[$checksum];
--- a/library/Director/IcingaConfig/IcingaConfigHelper.php
+++ b/library/Director/IcingaConfig/IcingaConfigHelper.php
@@ -74,7 +74,7 @@ class IcingaConfigHelper
 
     // TODO: Find out how to allow multiline {{{...}}} strings.
     //       Parameter? Dedicated method? Always if \n is found?
-    public static function renderString($string)
+    public static function renderString($string, $renderExpressions = false)
     {
         $special = array(
             '/\\\/',
@@ -100,7 +100,11 @@ class IcingaConfigHelper
 
         $string = preg_replace($special, $replace, $string);
 
-        return '"' . self::renderVariablesAsExpression($string) . '"';
+        if ($renderExpressions) {
+            return self::renderStringWithVariables($string);
+        } else {
+            return '"' . $string . '"';
+        }
     }
 
     public static function renderDictionaryKey($key)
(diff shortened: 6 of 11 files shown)

### CODE REVIEW

**Comment by Thomas-Gelf** [comment]
Nice one, thank you. What about just using $host.var.something$ instead of $$host.var.something$$ instead? I find $$ confusing, as $$ is what you use to escape dollar signs in runtime macros. Any idea how we could also allow for constants?

I also have objections against making this the default behavior for renderString. What about providing it only in a specific set of properties and also there only when rendering an apply rule?

**Comment by ardeois** [comment]
Sure I can change the `$$`convention, but I don't think using `$host.var.something$` is a good idea as it's the macro convention for icinga2. I think it would be nice that director doesn't override icinga2 macro and let them as is. We could use whatever we choose better, like `${something}`, `{something}`, `@something@` etc (Main issue would be retrocompatibility with existing values.)

Can you be more specific about allowing constants? I'm not very experienced with Icinga so I don't see what you're talking about.

Finally, I understand having this as default behavior in `renderString` could be dangerous, I can change it so only properties and custom properties could have these "variables". The first implementation of this PR was actually creating a `CustomVariableExpression` for "vars" properties and with overriding "renderObjectProperty" for default properties.
I could add this behavior back instead, however I think it's a shame to only use this feature for apply rules rendering. Technically, this could be used by anybody that want to render an expression using a variable. (e.g Host template `display_name = "Host: $$address$$`)

What do  you think?

**Comment by Thomas-Gelf** [comment]
> Sure I can change the $$convention, but I don't think using $host.var.something$ is a good idea as it's the macro convention for icinga2
Well, for me that's the reason for doing so :D People are used to this, so why shouldn't we use it? In custom vars it's already perfectly legal to do so, as they are evaluated at runtime. With this feature we would use the same syntax for properties that are evaluated only at config parse time. Icinga 2 doesn't allow to do so, therefore there is "bla" + bla + "bla". Why should a Director user even care about this difference?

> Can you be more specific about allowing constants?

```
PluginDir = "/usr/local/whatever"
MySpecialPrefix = "Special Service: "
object Service "whatever" {
  display_name = MySpecialPrefix + "whatever"
```

However, forget about this.

> I can change it so only properties and custom properties could have these "variables"

No need to support this in custom properties, as using $host.vars.whatever$ should already work fine there. It is supported by Icinga itself and evaluated at runtime. Start with something like this in IcingaService.php:

```php
public function renderDisplay_name()
{
    if ($this->isApplyRule()) {
        /// default behavior
    }

    return c::renderKeyValue(
        'display_name',
        c::renderStringAllowingMacros($this->display_name)
    );
}
```

Another approach could be in IcingaObject.php. Search for renderObjectProperty, it ends with:

```php
return c::renderKeyValue($key, c::renderString($value));
```
Replace it with:
```php
if ($this->isApplyRule()) {
    return c::renderKeyValue($key, c::renderStringAllowingMacros($value));
} else {
    return c::renderKeyValue($key, c::renderString($value));
}
```

The `renderStringAllowingMacros` should be added to the ConfigHelper of course. "Macro" is the Icinga 2 term for that kind of "variables".

ardeois:
> Well, for me that's the reason for doing so :D People are used to this, so why shouldn't we use it? In custom vars it's already perfectly legal to do so, as they are evaluated at runtime. With this feature we would use the same syntax for properties that are evaluated only at config parse time. Icinga 2 doesn't allow to do so, therefore there is "bla" + bla + "bla". Why should a Director user even care about this difference?

Well some differences I see between replacing macros ourselves rather than using native macro are:
 - "macro" resolves the variable name in the following order: User object, Service object,  Host object, Command object, Global custom attributes in the Vars ([See reference](http://docs.icinga.org/icinga2/latest/doc/module/icinga2/chapter/monitoring-basics#macro-evaluation-order)). That means, with macro, using `$something$` resolves to `host.vars.something`
 - With "expression", using `something + "whatever"`, it does not resolve to `host.vars.something`. Instead, a deployment error occurs: `Tried to access undefined script variable 'something'`
 - Using a "macro" in an apply service custom variable will show up in Icinga Web2 Monitoring like so:

 
 - However, using an "expression" in an apply service custom variable will show up in Icinga Web 2 Monitoring like so:

 

My concern with using the same convention in Director as is used in Icinga2 macros, is that it will behave differently and will be confusing for Director users.

However, after playing around a bit, using "expressions" such as : `host.vars.something + "whatever"`, I found that it only works in Apply objects. On this point, I agree that the feature should only work for "applyRules" objects.

Thanks for your feedback!

ardeois:
Hello @Thomas-Gelf,

Any update about this? I'd like to know what you think about my last comment. If you don't like it we could maybe use `$something$` but I think this should only be replaced specifically for apply for rules, and can't be used anywhere else as the "macro" behavior will be different if we replace all `$something$` ourselves.
I would prefer to support expressions in all objects though (or at least properties and custom variables)

msmol:
:+1:

Thomas-Gelf:
Hi @ardeois,

I'd still opt for $var$. Main motivation is that $$ is used to escape $ wherever the normal Icinga 2 macro resolver applies. So while something "new" like ${var} might work I consider $$ pretty confusing here. But also this would mean one more variable placeholder syntax. We already have too many of them.

In my believes for most users it might be easier to grasp that because of some "magic" $var$ in Director works in more places than in plain Icinga 2.

Biggest issue with this pull request to me is that I guess it currently renders what is expected, but it wouldn't do what one might think. Try to run service3.out from your tests with Icinga 2. I guess that a) it wouĺdn't validate and b) when fixed it wouldn't adjust the display_name as one might expect.
Honestly, I didn't try it by myself - it's just a blind guess. Problem is that parse order matters.

ardeois:
Alright I agree, even if it's a little different to replace $var$ ourself, most of the users won't see it so let's try it that way.
I'll change this PR to use $var$ as soon as I find a nice way to handle $$ escaping correctly.

Thomas-Gelf:
Cool! Any thoughts regarding the "I guess it has no effect" problem mentioned above? I still didn't try it by myself, so please feel free to prove me wrong in case I am:

> Biggest issue with this pull request to me is that I guess it currently renders what is expected, but it wouldn't do what one might think. Try to run service3.out from your tests with Icinga 2. I guess that a) it wouĺdn't validate and b) when fixed it wouldn't adjust the display_name as one might expect.
> Honestly, I didn't try it by myself - it's just a blind guess. Problem is that parse order matters.

Cheers,
Thomas

ardeois:
Yes I just tried processing service3.out, well an equivalent.
I've set up a host like this:


A service like this: (I've added `display_name` through API as it's not available through UI):


And it's process correctly in Icinga:


See the service name on top contains the `replace` var from host. And custom vars are properly displayed too.
For this example I didn't use `apply_for` rules as another PR handles that, so I've added an apply service with a dummy assign where

Thomas-Gelf:
> Yes I just tried processing service3.out, well an equivalent.

Ah, I see - you modified service3.out. So, just to be clear: this works for host vars, but will not for service vars for reasons mentioned above. Is this fine for you? Should we add some notes explicitly mentioning this?

ardeois:
Yes good point, add specific notes for that would be helpful. Should I add a .md file in `docs` or any place else?

Thomas-Gelf:
Good question :D Please open an issue on dev.icinga.org as a reminder, it isn't that urgent at all... what about compatibility? Would it be wise to add a schema migration file transforming $ to $$ in all affected properties to not break existing config?

Thomas-Gelf:
(hint: I'm off for at least the next three hours I guess - German Autobahn :p )

ardeois:
Yes I was wondering if a migration would be useful, but I think it's not. Right now only `$xxx$` will be replaced by this new code. A single `$` won't be replaced so this won't break.
Only variables that were interpreted as macro (by icinga core) would now be handled by director. So if we assume macros were already used like `$host.vars.custom_var$`, the change will be transparent to users.

Thomas-Gelf:
> BTW I ran the test of director and one is failing right now.

@ardeois: please pull/rebase again, should be fixed.

> Yes I was wondering if a migration would be useful, but I think it's not.

You're right, even better :)

ardeois:
Updated, I confirm all tests are green now, thanks

ardeois:
Hi @Thomas-Gelf,

So are you fine with this PR? do you think anything is missing or is it ready to merge?

Thanks,

Corentin

Thomas-Gelf:
Hi @ardeois,

please allow me to place some more questions that popped up while reflecting about this. It's basically "which properties should use this"? You mentioned display_name as a use case. Personally I hate that example as in my believes display_name shouldn't even exist for services. Don't want to repeat myself, you can have a look at this discussion to get an idea about my thoughts:

https://dev.icinga.org/issues/12497

Do you have a standalone use-case for this or would it mostly if not always only be used together with #23? And if so, why do you want to work with display_name? Wouldn't it be better to adjust the "real" service name and completely forget about display_name? And: any other properties where this could be helpful?

Please don't get me wrong, I do not question the feature itself, I'm just trying to figure out whether a) we are missing an important part (service name, especially in combination with #23) and b) whether it wouldn't be better to provide a "white-list" of properties where this should be applied.

So, I personally would want to use this solely for service object names, which is currently not supported in this pull request. So we should IMO try to extend it. But when it goes to individual properties... well, I found no example where I would use this. Do you have one?

Thomas-Gelf:
@ardeois: one nore thing. Custom variables. As we have seen, this isn't very helpful to access service variables within the service, it might be used to use host vars in service vars. What are your use-cases here? Is it to override service vars at a host level? Or what else? Could you show me some real-world example of where you would use this for when available in every custom variable?

I'm asking because there is a nice related feature in the master, I had to implement it for a customer. Documentation will follow, what it basically does is allowing one to override any service property at a per-host level. There are different variants of this:

* simplest one are services assigned to host templates. They are implicit assign rules and easy for me to find. When you click on a single host inheriting them you see them when going to it's services. New is that they right now allow you to have their vars/fields modified in-place. All this is stored in the host and will be deployed in a "magic" way
* what will follow are all applied services. Same story, as soon as the "show all services applied to this host" feature is done they will automagically work the same way
* the customer I work with uses apply rules but also syncs thousands of individual Puppet-generated services. Problem: it makes no fun to adjust individual thresholds in Hiera. So what they do now is: they set the "use_var_overrides" property to true ("y") in their sync rule. This feature is intentionally not available in the GUI but only through Sync/REST/CLI as it makes no sense without automation. But with automation it is pretty cool, as it allows you to sync whatever you want while allowing your users to override thresholds or other settings in a controlled way per host

So, this (overriding applied/imported service properties) on a per-host level is one big use-case for such custom variable magic and should now be covered in a very comfortable way. What other use-cases do you have for "magic expressions" in custom variables?

ardeois:
Sorry I don't have much time to answer today, but I'll take time to read your comments later and answer your this week end.

ardeois:
So for `display_name`, as I mostly agree with your comments in the feature, I feel like it's useful for generic apply service like Tcp checks or Process checks. Seeing a service in Icinga Monitoring with name `Apache2 Service` (that was generated from an apply_for where `host.vars.processes` is an array defining all processes to monitor) seems quite useful.
However I'm not strongly opinionated on that, if there is an easy way to see a service down in Monitoring module and see which service it is I'm fine with that.

> Do you have a standalone use-case for this or would it mostly if not always only be used together with #23? 
Right now no, I don't have any other use case than using expressions with #23. I was thinking this feature could be generic so whoever wants to use a variables from constant or whatever could do it. But once again, I'm not an expert to Icinga Director neither Icinga, so if you think expressions should only be rendered when an `apply_for` is selected, our use case would work. We first started to handle expressions for everything, but it seems like it's not working for every object, so I restricted it to Apply objects.
If there is a white list and we are able to edit it (meaning I have in my apply service a custom var `tcp_port` that will be an expressions `$config$` retrieved from the apply for, this can work. But is this necessary though? Do you think not having a white list is dangerous?

As for our use case, it seems similar to your example, on the difference that we will have 3 types of users in Icinga Web. But before talking about that, I think I need to explain what we're trying to achieve.
We will use Icinga platform to monitor our customers machines, but we'll manage everything for them. There will be different set of monitoring (tcp ports, processes, pings etc).

Then we'll have 3 groups of people using Icinga Web (and director)
 * Group 1: Will create Service Templates, Host templates, zones, generic apply for rules etc (through UI or API). So they will handle the platform and the automation
 * Group 2: Configure services for a client's host (pre-added through API or sync rule, we're not sure yet). So we want them to be able to ONLY touch a host, where they'll define in custom variables which tcp port, which process, etc. Those users are not very technical, so the interface should be as simple as possible and restricted (with permissions) to host only (custom vars is even better but host permission is enough). So the easier the better.
* Group 3: Monitor alerts on services and act on it if something is wrong. So they want to see easily what's going on (though Icinga Monitor module) and be able to know what service is down (Tcp 80, or Process Apache2 is down). I think this group won't need to access Director though.

So let's take an example of one host (web.example.org) that was created automatically. The customer service will configure the `hosts.vars.processes` array in the UI to let's say 2 dictionaries containing for each item:
```
{
   snmp_process_name: "VBoxService"
   snmp_warn: "1.5",
   snmp_crit: "0,5"
   display_name: "Virtual Box"
}
```
```
{
   snmp_process_name: "ruby"
   snmp_warn: "0",
   snmp_crit: "0,4"
   display_name: "Web Api"
}
```
All of theses keys in the dictionary are already defined and can't be changes, validation of string and number for each key is necessary, to ensure no mistake was made.

Once this is saved, we want it to be deployed automatically (through job sync I think), and we want to see 2 services generated in Icinga, corresponding to those 2 processes.
We want the flow of adding a new Process for "Group 2" as easy as possible, and as controlled as possible (controlled from mistakes) as they are not IT guys.
We want "Group 3" to understand clearly what is down, for example I feel like seeing `Service Virtual Box` down is easier to look in the dashboard than `Service Process check` (that would be the generic service template name.

I don't know if I'm making myself clear, it's friday afternoon for me (even later for you), but I hope that would help you see what we're trying to achieve.
However we mostly discovered Icinga, Icinga Web and Icinga Director few weeks ago, so if you have any different approach or suggestion, we're all ears.

Thanks and enjoy your bottle of wine :)

Corentin

Thomas-Gelf:
Thank you Corentin,

that helped a lot to improve my picture of what you want to build. Lets go more into details next week, but let me already say that I think that a lot of what you want to achieve should already be possible with what I pushed to master this week. Have a look at those screenshots, let's start with a service inherited from another host template:

director_override_inherited_service

Next, there is an imported service, a single host's local service. Normally, when you would modify it the next sync would destroy those properties. So the use_var_overrides flag set on import has this effect:

director_override_imported_service

The service will not be modified when the user adjusts those vars, this is what the preview looks like:

director_override_preview

There is also a related "magic" apply for implementation, however this is still a hidden feature and I'm currently unsure whether I want to keep it. Services generated that way offer the same "vars override" mechanism:

director_override_magic_applyfor

All of this is to be found in the current master since yesterday, I guess parts of this could be used for some of your requirements. Still, this doesn't mean that we shouldn't try to merge those feature-requests.

> Thanks and enjoy your bottle of wine :)

I did, now it's empty :-p

Have a nice Weekend!
Thomas

Thomas-Gelf:
@ardeois: one more screenshot, this is what Director currently renders for those features. DirectorOverrideTemplate is always rendered, the "magic apply for" is in the master since a long time but hidden and only rendered if at least one host makes use of it. Please note that this is not a replacement for individual "apply for" rules:

director_renered_magic

It has initially (quite some time ago) been built to satisfy the "I want to import services on hosts and still be able to override their vars" requirement. As the new more generic vars-override-mechanism is much more powerful I'm currently evaluating to silently drop that magic-apply-for feature as I see not much use for it anymore, since yesterday it's mostly obsolete.

ardeois:
Hi @Thomas-Gelf,

Thanks for your detailed feedback, some of these features could be useful indeed, especially the vars override.
However right now I think the main feature we miss to achieve our goal is the "apply for" rule on services. Most of our hosts (or host template) won't have predefined services, but essentially need a list of services depending on some vars. For example, a host will have a list of "tcp_ports" in its vars and a "Tcp Service" for each item should be generated.
As I see a service inherited from host template will be useful for some use cases (maybe some of ours too), I'm not sure this is applicable for this case, as we have an unknown list of services to generate from an array, am I right?

As for the magic apply for, I did see the implementation in the code and honestly I didn't understand how it worked and was confused on the purpose of it. Now I understand and I agree with you this feature becomes obsolete.

So this said, I still need the apply_for PR in director, which depends on this PR, being the first step. I'm not sure this feature could be done with the current master, but maybe I missed something on your explanation?
We won't be able to develop in Director the next 2 weeks but we'll still try to answer and address any concern you have on those PRs. I rebased them and will try to keep them up to date with master as much as I can.

Thanks,

Corentin

test/php/library/Director/Objects/IcingaServiceTest.php
```diff
@@ -12,6 +12,7 @@ class IcingaServiceTest extends BaseTestCase
     protected $testHostName = '___TEST___host';
 
     protected $testServiceName = '___TEST___service';
+    protected $createdServices = [];
```
Thomas-Gelf: We also hate it, but we force ourselves to stay compatible with PHP 5.3 as there are still "Enterprise" Linuxes shipping with that one. Please use array() instead of [] here (and in other places).

Thanks,
Thomas

Thomas-Gelf:
After some minor modifications I merged this into my local master, but didn't push it yet. The branch feature/nested-apply-rules-12033 has already been rebased on this, so you could give a look at it over there. Once I merged that one too I'll push the master upstream. Hopefully within today, PDT.

It now also allows for empty object names and expression in the object name. The latter one would result in an additional name property when rendering the service.

Closing this pull request, thanks for your patience and the good work!

PR #23, opened 2016-08-31
Add support for Apply for rules in services

ardeois:
Apply for rules are defined through `apply_for` property
Only Array custom variables will be eligible in apply_for dropdown.

API Example:
```bash
 ./director-curl POST director/service?name=my_service '{"apply_for": "for (checked_tcp_port in host.vars.checked_tcp_ports)" }'
 ```

 Will render as:
 ```
  apply Service "my_service" for (checked_tcp_port in host.vars.checked_tcp_ports) {
      ...
  }
  ```
Feature: https://dev.icinga.org/issues/11976

Depends-On: refs #11976 (https://github.com/Icinga/icingaweb2-module-director/pull/20)
refs #11976

### CODE REVIEW

**Comment by ardeois** [comment]
Here is a how it looks like in Service apply-rule configuration:

**Comment by msmol** [comment]
Another :+1:

**Comment by Thomas-Gelf** [comment]
Hi @ardeois,

no need to implement all possible variants right now, however I'd love to get clear ideas on some details before merging this. Could you please let me know your thought regarding use-cases as the following:


Please give me some more time for this. There are various related
use-cases, like the following:

apply Service "HTTP Check " for (domain => config in host.vars.vhosts) {
  vars.port = 80
  vars += config
}

How would you tackle this? This reads as "set default port 80 in the apply rule but let the host var win if any". Should we always accept all host vars? With the Dictionary data types we could enforce a specific structure, so this should be fine.

I see no conflict with this and your pull request, just curious about your thoughts.

Thanks,
Thomas

**Comment by ardeois** [comment]
Well indeed we thought about that too and our idea (even not ideal) would be to tackle the `var += config` in a different way. As we now `config` is a dictionary and all the keys are defined we could do something like this:
```
apply Service "HTTP Check " for (config in host.vars.vhosts) {
vars.port = config.port
vars.something = config.something
}
```

It's not perfect but that would work. I feel like handling `vars += config` in director UI will be complicated and will introduce unknown in what this `config` object contains.

Notice also that our PR is not doing `for (domain => config in ...)` but `for (config in ...)`. The main reason is that the first one only works with dictionaries, the second only with arrays. As handling both would have introduce too much complexity, we think it's much easier to handle only arrays in apply for rules. This also simplifies the UI for future Array of Dictionaries as I don't see an easy way to create a UI for Dictionary of Dictionaries. For example:
```javascript
host.vars.processes = {
  "Apache Service" : /* key is generic here so it can't complicated to define the dictionary structure*/ {
      "snmp_process": "apache2"
      "critical_level": 1,
  },
  ...
}
```
Would be transformed as:
```
host.vars.processes = [
 { /* Dictionary is well defined and keys are always the same */
      "display_name": "Apache Service",
      "snmp_process": "apache2"
      "critical_level": 1,
  }
]
```

friesoft:
Any chance to get this merged soon? :)

Thomas-Gelf:
@friesoft: I'm allowed to dedicate quite some to Director during the next weeks and I'd love to have this merged before 1.2, scheduled for the beginning of November. Current show-stopper here is that Dictionary handling is still unspecified, once we figured out how to deal with them we should be ready to merge this.

Thomas-Gelf:
@friesoft: I'm allowed to dedicate quite some to Director during the next weeks and I'd love to have this merged before 1.2, scheduled for the beginning of November. Current show-stopper here is that Dictionary handling is still unspecified, once we figured out how to deal with them we should be ready to merge this.

Thomas-Gelf:
@friesoft: I'm allowed to dedicate quite some to Director during the next weeks and I'd love to have this merged before 1.2, scheduled for the beginning of November. Current show-stopper here is that Dictionary handling is still unspecified, once we figured out how to deal with them we should be ready to merge this.

Thomas-Gelf:
@ardeois: had some time to wrap my head around this. Know I feel pretty confident that there is nothing that would really hurt in future, especially as there is nothing what speaks against adding key => config later without breaking anything existing.

I prepared feature/assign-for-11976 with your changes. If it is OK for you I'd clean some little things up and merge it to master - eventually within today.

Cheers,
Thomas

Thomas-Gelf:
> Would be transformed as:
> 
> host.vars.processes = [
>  { /* Dictionary is well defined and keys are always the same */
>       "display_name": "Apache Service",
>       "snmp_process": "apache2"
>       "critical_level": 1,
>   }
> ]

This wouldn't have worked. It would use each entry serialized (multiline) as part of the service name, pretty scary. To fix this I added the possibility to use variables in your service name. So in your example I would use "Process check: $config.display_name$" and get the following result rendered:

Would be rendered as:
```
apply Service for (config in host.vars.test1) {
    name = "Process check " + config.display_name
    ...
```
I'm also allowing for empty service names, that way you get only "config" as a name and that's it. Guess the form should get better explanation, eventually I'll restructure some things there but this will happen after I merged the nested apply rules. For this feature here I'm considering to add two more flags, to allow one to find alternative variable names instead of "config". Given all this we shouldn't be that far from Dictionary-support in "apply for".

I have this pull request and #20 together in my local master. Will now sleep a little bit and hopefully push it later on this day.

Cheers,
Thomas

Thomas-Gelf:
The branch feature/nested-apply-rules-12033 has already been rebased on my local master where I already merged this, so you could give a look at it over there. Once I merged that one too I'll push the master upstream. Hopefully within today, PDT.

Please note that your example with the "Array of Dictionaries" would produce something scary. You'd be left with services containing a Dictionary rendered to a multiline string making part of your object name, looking funny at best in the web frontend and causing errors when submitting commands to them. Therefore I added the possibility to use expressions also in object names to the related pull request, that would allow one to smoothly work around this. I'd love to also see support for "key => value in whatever" anytime soon, but this isn't that urgent.

Closing this pull request, thanks for your patience and the good work!

PR #24, opened 2016-09-01
Add Support for dictionary custom fields

pboutin:
After some discussions with @Thomas-Gelf for [feature #12093](https://dev.icinga.org/issues/12093), my team from Internap and me, came to this first implementation.

You can define your dictionary structures in the added "Dictionaries" tab, within the custom field section. Once the dictionary is created, you can specify its fields. It's possible to define sub-dictionary by creating a dictionary field. Here's the UI for that : 
screenshot

Now that the dictionary is created, you can use it as a custom field. For now, the dictionary is presented into a JSON structure that you can edit.

We wanted to provide some validations in order to preserve the dictionary's good structure. Here's the rules :
- All keys must be there, at any time
- Fields must respect the type of the structure (you can't put an integer if the field expect a string)
- If a field is marked as "required", it can't be `null`

screenshot

Finally, the dictionary can be applied to an entity, a host for example.
screenshot

This PR also ships with automated testing.

Also, we will produce the according documentation when the implementation and design details will be validated and accepted.


Thomas-Gelf:
Great, that's already very close to what I wanted to see, perfect, well done! There are some little things I disagree with, but nothing really important at all. There are also some merges in this branch I would prefer to not see in the master tree for better readability.

However, to not loose too much time discussing such little issues across timezones (especially when not really relevant to the feature at all) I'd like to ask whether it is ok for you when I cherry-pick those parts that I consider ready-to-be-merged, adding slight modifications where required shamelessly in your name?!

Then we could rebase the remaining parts and discuss what's eventually still missing.

Thomas-Gelf:
Hi @pboutin,

I love this pull request :-) A full rebase/squash wouldn't have been necessary, but it's also fine of course. Issues so far: a poop description, incomplete PostgreSQL migration and missing changes to the base schema files. Please let me know in case you need any help with the latter.

Still needs nested form element rendering, something I promised to help with. In the meantime I'd like to merge most if not all of this to master as soon as SQL migrations are fine.

Cheers,
Thomas

pboutin:
@Thomas-Gelf : I fixed the issues you pointed out and I updated our branch so it can be up to date with your master. If you have any concerns, let me know in this PR.

Thomas-Gelf:
Hi @pboutin,

could you please always rebase your branch on the current master? This would make it easier for me to follow your changes, merges forth and back are sometimes incredibly hard to grasp at first sight.

One thing I missed: there first was the pull request by @msmol with this schema:

https://github.com/internap/icingaweb2-module-director/pull/1/files#diff-03bcf9230662db87df79064a3248331d

Now we have this here:

https://github.com/Icinga/icingaweb2-module-director/pull/24/files#diff-4f27aca0ba446bf5f6c8573df29b530d

This leads to lot's of duplicated logic and is contrary to what I expected to see after the discussion here:

https://dev.icinga.org/issues/12093

Didn't realize this earlier sorry - I somehow confused those two similar pull requests. What was the reason for this? Did you have problems with the other approach, was there a dead-end with a problem we didn't see at the beginning?

Thanks,
Thomas

ardeois:
Well I can answer to that. We first did the development like you suggested, it was working but the UX was really not intuitive. To create a dictionary field (containing a sub dictionary) like the example bellow:
```json
{
   key1: 0,
   key2: {
      sub_key1: ''
   }
}
```
The steps were the following:
1. Create a Datafield, `subkey` of type String
2. Create a Dictionary of name `Sub dictionary`
3. Create a DictionaryField associating `Sub dictionary` and `subkey`
4. Create a datafield `key1` of type Number
5. Create a datafield `key2` of type Dictionary and select `Sub dictionary`
6. Create a Dictionary `Main Dictionary`
7. Create a DictionaryField associating `Main Dictionary` and `key1`
8. Create a DictionaryField associating `Main Dictionary` and `key2`
9. Finally you can create you Datafield of type Dictionary and select `Main Dictionary`. This could now be used in hosts.

As you can see we had to create the deepest element first, then go up to `Main Dictionary`, this was not very intuitive and while playing with it we always confused between what to create first, DictionaryField, Dictionary etc.

So with the new implementation the steps are:
1. Create a Dictionary `Sub dictionary`
2. Add in it a field `sub_key1` of type String
3. Create a Dictionary `Main Dictionary`
4. Add in it a field `key1` of type Number
5. Add in it a field `key2` of type Dictionary and select Main Dictionary
6. Finally create a Datafield of type Dictionary and select `Main Dictionary`.

This implementation takes really less steps and and seems more logical to me. Now I agree we re-implemented our table DictionaryField similar to DataField, but the code is inherited from DataField, to share the most code as possible. Your proposition was I think the best solution in term of code reutilisation but was tricky. This one is better in term of user experience, but had to compromise on some code duplication.

Let us know your opinion about that.

Thanks,

Thomas-Gelf:
Hi @ardeois,

this is still not clear to me, as IMO there is no reason why you couldn't just use new forms to fill the datafield tables in a slightly different / shorthand way if UX where the "only" concerns. What I miss in the current implementation is flexibility. I want to be able to also use special / restricted data types like a "data list" field or types provided by foreign modules in a dictionary. Seems that this is currently not possible.

But don't worry, this can somehow be solved. I'll start right now to tackle it from the other side: write some code rendering a related nested form element (to replace the current text area) to getter a better feeling of all this.

To get another related question answered I'd like to go an indirect way through an example:

```
object Host "webserver" {
  vars.vhosts = {
    "example.com" = {
      domain = "example.com",
      aliases = [ "www.example.com", "www2.example.com" ],
    },
    "something.local" = {
      domain = "something.local"
    }
  }
  vars.tomcat_hosts = {
    "example.com" = {
      domain = "example.com",
      aliases = [ "www.example.com", "www2.example.com" ],
    },
    "something.local" = {
      domain = "something.local"
    }
  }
  vars.list_of_hosts = [{
      domain = "example.com",
      aliases = [ "www.example.com", "www2.example.com" ],
  }, {
      domain = "something.local"
  }]
}
```

The inner properties are simple. A "Single Vhost" is defined as follows:

```json
{
  "domain": {
    "datatype": "string",
    "required": true
  },
  "aliases": {
    "datatype": "array",
    "required": false
}
```

What properties do you use to specify "Multiple Vhosts" used in "vhosts" and "tomcat_hosts" and how do you declare "List of Vhosts" used in "list_of_hosts"? In the meantime I'll bring my kids to bed and open a bottle of wine for better inspiration ;-)

Cheers,
Thomas

ardeois:
Hi @Thomas-Gelf,

>  I want to be able to also use special / restricted data types like a "data list" field or types provided by foreign modules in a dictionary. Seems that this is currently not possible.

Well that's not entirely true, you can select any type for a dictionary type. This interface is actually directly from DataTypeForm:


However you are right that in the Host object you don't inherit the UI from customs DataType. That's something we wanted to tackle in a phase 2, mostly because we don't know how to do that with Zend.

>What properties do you use to specify "Multiple Vhosts" used in "vhosts" and "tomcat_hosts" and how do you declare "List of Vhosts" used in "list_of_hosts"? In the meantime I'll bring my kids to bed and open a bottle of wine for better inspiration ;-)

That's exactly what we are working on right now, first we wanted base dictionaries working (with defined structure), now we want to work on `Array of DataType`, where you can specify which datatype the array is. In this case that would be `Array of Dictionary`.
While this is not done yet, and honestly we have no idea how to do it yet (so ideas are appreciated), we think creating the base DictionaryType is a feature by itself. Then creating a new dataType to handle Array of generic types and handle a proper UI with it will achieve your (and our) goal.

So both `Array of Type` and `DictionaryType` will be needed for your example, but the data would be structured like `list_of_hosts`:
```
object Host "webserver" {
  vars.list_of_hosts = [{
      domain = "example.com",
      aliases = [ "www.example.com", "www2.example.com" ]
  }, {
      domain = "something.local"
  }]
}
```

Best,

Corentin

Thomas-Gelf:
Great, this takes the right direction. And I got my bottle of wine, so things can only get better now ;-) Array of Type is fine for me, that's what we had in the discussion in the initial ticket. I was trying to understand whether your plans are still going in that direction. Implementation: I'd add a DataType allowing you to choose another type and proxying that types settings afterwards. Please let me know if you need any help with this.

I made the vhost VS tomcat_host example to point out a specific problem and to make clear that the property name is NOT part of the dictionary definition. In the current implementation allow_multiple suggests to be the "Array of Type", so things are not so clear here yet. We could misuse allow_multiple or a similarly named property for the first part of the vhosts example. Or introduce one more type. What about Dictionary and "Hash of Something" with a "Hash of String" eventually being also offered as an "Enum" shortcut? And should our current dictionary then still be called Dictionary, as this is more or less an alias of Hash. Isn't it something like a "Defined Type"?

The wine is good :)

ardeois:
Well I'm not sure to understand 100% of your point, but `allow_multiple` right now is not implemented and I think it should be removed as this could be tackled with `Array of Type`.
As for your idea about hash, do you mean handling objects like this?
```
vars.vhosts = {
    "Custom Name" = { //Custom name being a custom value the user can enter
      domain = "example.com",
      aliases = [ "www.example.com", "www2.example.com" ],
    },
```

The first definition of our objects was similar to this, but after some though I feel like it's complicated to have a UI proposing this, so we settled to always use arrays of dictionaries, where each dictionary has its keys defined. I mean `Custom Name` could be defined in the dictionary with its own key like:
```
{ 
   domain = "example.com",
   aliases = [ "www.example.com", "www2.example.com" ],
   name = "Custom Name"
 }
```

Thanks for all your feedback

Thomas-Gelf:
> Well I'm not sure to understand 100% of your point, but allow_multiple right now is not implemented and I think it should be removed as this could be tackled with Array of Type.

That was the point :)

> As for your idea about hash, do you mean handling objects like [..]

Exactly. Because this is what for example makes part of (or full) object name in apply for rules with dictionaries, that's what they are mostly used for.

> [..] complicated to have a UI proposing this [..]

I'm sure we'll find a solution. Solving this is essential for me for this feature, I see them pretty often. Sometimes it's just because of the datasource shipping them this way, otherwise it's usually because of "apply for".

> Thanks for all your feedback

I hope it's not too much. Please let me know before getting angry with me ;-)

ardeois:
Don't worry I'm not angry at you at all, on the contrary I really appreciate your feedback. Having discussions back and forth is part of the deal when you contribute open source.

> [..] allow_multiple right now is not implemented and I think it should be removed as this could be tackled with Array of Type.
>> That was the point :)

So we thought about implementing `allow_multiple` (in a second phase/PR) but if we do implement it like that, that would mean it's possible to have "Arrays" of dictionaries only. While this work with most use cases, it's kind of sad we can't benefit "Array" of anything in a DataField  or a Dictionary directly right?
That's why we though of tackling the problem differently by creating a new DataType `Array of ...`, this would address the use case of having multiple items of a dictionary. And it will give you the ability to create `Array of` number, boolean or whatever DataType is defined.
While more complex it's also more generic as we just implement a new DataType. If you do agree with this approach, we will however need some help (or example) on how to handle the UI, the main issue being DataType implements `getFormElement` that need to return a `ZendElement`, but in our case it will be a subform I guess.

>As for your idea about hash [..]
>> [..] Solving this is essential for me for this feature, I see them pretty often. [..]

I would definitely like this feature as well, but we found a workaround if we implement array of dictionaries instead of hash. Both looks complicated to implement in the UI but the former seems a little bit easier. However I think both features could coexist in director, but we need to separate them, like you said, one being a Dictionary only (or "Array Of" "Dictionary"), the other one being "Hash Of" "Dictionary"

plarivee:
@Thomas-Gelf any news on when this will be merged / rewritten? Would really like to be able to use Director 1.2 and move on from ( https://github.com/internap/icingaweb2-module-director ) forks

Tontonitch:
This is also a feature that I'm waiting for very impatiently.
It is the major blocking point in our future migration to Director.
Regards,
Yannick

cbnorman:
@thomas-gelf is there any plan to impliment this. I’m now using data lists which has been very useful for apply services to interfaces. Would be nice to apply extra variables to each switch interface.

plarivee:
I would close this and better do a feature request. All people involve in this do not work anymore for the 
 company involve in the original pull request.

PR #26, opened 2016-09-05
adds systemd script as contrib

dgoetz:
Expands the systemd example from the docs and adds it in contrib for simplifying the setup.


Thomas-Gelf:
Committed a slightly modified version in your name, thanks! Please always check for trailing whitespaces and try to fold markdown at about 80 characters to improve readability in plaintext editors.

Cheers,
Thomas

PR #27, opened 2016-09-12
Include the Icinga 2 Windows plugins by default in powershell script.

mengine23:
Im not sure if this was wanted by the author, but i am not possbile to execute windows check commands (i.e. service-windows) on my client without this include.


mengine23:
Fixed with 160a630

PR #31, opened 2016-09-30
Corrected a misleading string

konstantin-kelemen:
This corrects a message string on the Host/Agent page.


Thomas-Gelf:
Committed, thanks!

PR #33, opened 2016-10-24
Added support for array of any Datafield

pboutin:
Here's the first draft of the "Array of anything" datafield. The way I did it, is by using Datafields instead of Datatypes. By doing it like, we make it compatible with more complex types such as Dictionaries or Datalist.

To use it, you have to declare a "unit" field, like a `DataTypeNumber` for instance.
After that, you can declare your "meta-array" field by creating a `DataTypeMetaArray` field, and in the settings subform, you will be able to say "I want an array of `port (DataTypeNumber)`.

On the other side, when you actually use the field, it simply renders the fields using the proper form element.

Pros:
- it works well, even with our dictionaries
- fairly simple (4 files)

Cons:
- a bit hackish, I don't like to carry `fakeElement/subElement` all over the place just to validate stuff or render the form
-  not unit tested, waiting for inputs on that

### DIFF
--- a/library/Director/Web/Form/Element/MetaArray.php
+++ b/library/Director/Web/Form/Element/MetaArray.php
@@ -1,0 +1,62 @@
+<?php
+
+namespace Icinga\Module\Director\Web\Form\Element;
+
+class MetaArray extends FormElement
+{
+    public $helper = 'formMetaArray';
+
+    /**
+     * Always ignore this element
+     * @codingStandardsIgnoreStart
+     *
+     * @var boolean
+     */
+    protected $_ignore = true;
+    // @codingStandardsIgnoreEnd
+
+    public function isValid($values, $context = null)
+    {
+        if ($values === null) {
+            return true;
+        }
+
+        $subElement = $this->getAttrib('subElement');
+
+        foreach ($values as $value) {
+            if ( ! $subElement->isValid($value)) {
+                $this->addError(sprintf("'%s' is invalid", $value));
+                return false;
+            }
+        }
+        return true;
+    }
+
+    public function setValue($values) {
+        $filteredValues = [];
+        $subElement = $this->getAttrib('subElement');
+
+        foreach ($values as $subValue) {
+            $subValue = $subElement->setValue($subValue)->getValue();
+            if ($subValue) {
+                $filteredValues[] = $subValue;
+            }
+        }
+        return parent::setValue($filteredValues);
+    }
+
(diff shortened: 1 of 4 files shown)


Thomas-Gelf:
I like it. That fake/subelement is IMO the way to go as long as we depend on Zend_Form. What about a) hiding the fact that we are doing so by just calling it "Array" instead of "MetaArray"? When you read "MetaArray" you start wondering what kind of beast this could be, it scares you - while by the end it is just an "Array of Something". No one forces us to make our internal issues so obvious. So, b) as the name "Array" is already in use, why not making every array an "Array of Something" with "String" being the default when no Sub-Datafield has been chosen.

It's absolutely reasonable that you opted for Fields instead of types, no objection here. We could eventually evaluate whether we could either allow to also opt for one or the other. Or we could dedicate some more love to the DataFieldForm in relation with the Array, as it could automagically create sub-fields for us. Just to save some clicks. But let me play with it first to get clearer ideas about this first. I guess I'll then also come up with questions related to sort order and so, we'll see ;-)

pboutin:
Cool, glad you like it, I might continue to work on that during the week, feel free to play with it if you have the time. We would like to get it merged on our side in the next few days, if we could have it done quick it would be great.

For the Array thing, you suggest to move the logic into the existing Array DataType ?

library/Director/Web/Form/Element/MetaArray.php
```diff
@@ -0,0 +1,62 @@
...
+            if ( ! $subElement->isValid($value)) {
+                $this->addError(sprintf("'%s' is invalid", $value));
+                return false;
+            }
+        }
+        return true;
+    }
+
+    public function setValue($values) {
+        $filteredValues = [];
```
pboutin: Sorry, reflex :P  will fix in the next commits

Thomas-Gelf:
> For the Array thing, you suggest to move the logic into the existing Array DataType ?

Yes.

pboutin:
I like the idea of having only one Array datatype. Would be more intuitive. I took the time to investigate how the existing Array worked, with that ExtendableSet element. I'm not sure how we could merge these two togethers. That would require to change the helper we use according to the type of the array, but I've no clue how to achieve this huh.

plarivee:
@Thomas-Gelf  any news on when this will be merged / rewritten?  Would really like to be able to use Director 1.2 and move on from ( https://github.com/internap/icingaweb2-module-director ) forks

PR #37, opened 2016-11-02
Feature/enhance bash script agent rollout 12910

chrnie:
I've added an Agent install script for linux and modified the download link.
Updated windows agent ps1 script


chrnie:
rebased in a new repo

PR #40, opened 2016-11-25
PowerShell 2.x and Windows 32bit support

LordHepipud:
There was an issue with downloading the Icinga 2 Agent on PowerShell 2.x as the Invoke-WebRequest function does not exist there.

Furthermore there was a problem with the detection of the default install dir on 32bit Windows systems because of a wrong environment variable.


Thomas-Gelf:
Pushed to master, thanks

PR #42, opened 2016-12-01
quick fix for broken Notification Template tab

mdetrano:
Quick fix for missing Notification Template table object.


Thomas-Gelf:
Has been fixed otherwise, thank you anyways for this patch!

PR #783, opened 2017-02-13
Include plugins-contrib by default for Linux agents

dnsmichi:
This makes it easier to deploy CheckCommands out-of-the box where the Icinga Director already uses them from the master, but requiring them on the client as well.


dnsmichi:
Lazy users ... https://monitoring-portal.org/index.php?thread/39618-icinga2-nutzt-falsches-plugindir/&postID=242565#post242565

dnsmichi:
Obsolete, 66ff070c778a9ed5ccca8cf09c3fcdd4c43957f9 introduces this in Git master.

Issue #651, opened 2016-12-14
[dev.icinga.com #13559] Agend Configuration Script, make Variable ICINGA2_PARENT_ENDPOINTS an Array


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/13559

**Created by shoenscheid on 2016-12-14 09:17:05 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-12-14 09:17:05 +00:00 (in Redmine)_

---

To allow the configuration of Agents with Multiple Master nodes, the Variable ICINGA2\_PARENT\_ENDPOINTS should be able to take multiple values, an array would be a good type here.

./confagent.sh: line 10: icinga02d100.dev.example.com: command not found

check: icinga2 installed - OK

1 \#!/bin/bash
2
3 \# This generates and signs your required certificates. Please do not
4 \# forget to install the Icinga 2 package and your desired monitoring
5 \# plugins first:
6
7 ICINGA2\_NODENAME='icinga-client03d100.dev.example.com'
8 ICINGA2\_CA\_TICKET='55e96ca2d40d9c9be3ef59c5fcaeea70df3a1e7c'
9 ICINGA2\_PARENT\_ZONE='master'
10 ICINGA2\_PARENT\_ENDPOINTS='icinga01d100.dev.example.com', 'icinga02d100.dev.example.com'
11 ICINGA2\_CA\_NODE='icinga01d100.dev.example.com'
12 \# Make sure icinga2 is installed and running

mens:
I submitted a pull request #881 to fix this problem

Thomas-Gelf:
This should have been fixed with v1.5.0

PR #881, opened 2017-03-30
Support multiple parent zone endpoints in linux agent bash script

mens:
This patch adds support for satellite zones and parent zones with multiple masters to the linux agent icinga2Agent.bash script. This fixes issue #651

### DIFF
--- a/contrib/linux-agent-installer/Icinga2Agent.bash
+++ b/contrib/linux-agent-installer/Icinga2Agent.bash
@@ -85,10 +85,25 @@ ZONES_ICINGA2=`cat << EOF
 
 // TODO: improve establish connection handling
 object Endpoint "${ICINGA2_NODENAME}" {}
-object Endpoint "${ICINGA2_CA_NODE}" {}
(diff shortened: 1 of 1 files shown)


Thomas-Gelf:
Thanks for this pull-request @mens, and sorry for not merging it for such a long time. Problem for the delay was that it doesn't deal with the connection direction (Agent to Parent or the other way round), but I failed to even tell you my concerns. We should have merge it anyways, as it fixes an obvious bug.

As of today, I would prefer to not deal with all the potential variants (Endpoint object with/without host property) in this shell script and move the rendering logic into Director. Advantage: a single script for every host, ready for copy & paste. Disadvantage: people using this script as a base for their very own automation would need to implement the very same logic. But that's what they're forced also as of today I guess.

Thomas-Gelf:
Hint: please follow #1504 for related discussion.

lazyfrosch:
This should be fixed with #1596 - if there is any issue with the new script please open a new issue.

Issue #934, opened 2017-05-17
Use different DB tables for different object_types


Thomas-Gelf:
Currently for example we have single service objects, services being service set members, service templates and apply rules - all of them in the same table. This absolutely made sense, as it helped to reduce duplicate logic, code, schema and so on. But as we added more and more diverging features, this started to become a problem.

Working with separate tables would help to improve performance. And it would allow us to solve some issues like currently having no way to enforce service template name uniqueness.

Implementing this will require quite some refactoring.

PR #947, opened 2017-05-22
Sync, SyncRule: fix import of service templates...

### LINKED ISSUE #945: Syncing service templates and service apply rules does not work properly
When I sync service templates or service apply rules from an import source using "Trigger this sync", director always creates new ones; it does not notice it has already done that.

rattacresh:
...and apply rules

fixes #945


Thomas-Gelf:
Hi @rattacresh,

thanks for your report and the pull request. A clean fix for this would be #934, as it would bring back DB-backed uniqueness for all those Service variants. Currently, synchronizing templates should (in theory) work when setting object_type to template and using the object_name as key. In case it doesn't, IMO the fix should be to load only existing templates when object_type is set to `template` in the Sync Rule. I didn't test it, are you sure that this currently fails with templates once configured as mentioned?

I'll check the part with the Service Set, looks good at first sight.

Apply rules should in theory also be handled, but they are trickier. It is absolutely legal to have multiple apply rules carrying the same name. So, no chance to reference them from outside in a meaningful way.

rattacresh:
Yes, it fails even with templates once configured as mentioned: Without my patch, $this->rule->hasCombinedKey() in Sync::loadExistingObjects() will be false. Thus, IcingaObject::loadAllByType() will be called. Because IcingaService::keyName is array('host_id', 'service_set_id', 'object_name'), we have is_array($class::create()->getKeyName()) and so it calls $class::loadAll($db, $query), which in this case is DbObject::loadAll(). Since, by default, keyColumn === null, it simply appends the objects to the array ($objects[] = $obj). The key will hence be the array index. In most cases, this won't be equal to any object_name, and so director will assume that the object needs ot be created. If, however, object_name happens to be a number, then very strange things may happen.

You are right that my patch does not allow multiple apply rules with the same name. However, it is still better than the current situation and can be fixed later.

dmitriy-terzeman:
Is there any updates on this issue? i facing the same issues with director v1.8.0, Sync rules don't work correctly with service sets, service(when object_type is template) and apply rules.

PR #948, opened 2017-05-22
Sync: Allow syncing lists of template imports

### LINKED ISSUE #946: On sync, it should be possible to import a list of templates
Currently, syncing only supports importing a fixed number of templates. It should be possible to specify a list of templates to import.

rattacresh:
fixes #946

### DIFF
--- a/library/Director/Import/Sync.php
+++ b/library/Director/Import/Sync.php
@@ -430,7 +430,11 @@ class Sync
                         $newVars[$varName] = $val;
                     } else {
                         if ($prop === 'import') {
-                            $imports[] = $val;
+                            if (is_array($val)) {
+                                $imports = array_merge($imports, $val);
+                            } elseif (!is_null($val)) {
+                                $imports[] = $val;
+                            }
                         } else {
                             $newProps[$prop] = $val;
                         }


Thomas-Gelf:
Just to get a better understanding of what you are trying to achieve: is your problem that you have to combine multiple source columns with one of those being an array? Could you please show me a very few sample data rows from such an Import Source?

Also: order is important when it goes to imports. Once we have a related example I might ask you more related questions targeting merge order. Eventually array_merge might not be the best strategy here.

NB: And don't worry about the red test, this has been because of the phpcs 3.0 release, it dropped support for PHP 5.3. We will also do so very soon, but I forced Travis to stay with 2.9.x in the meantime. Just sqash & rebase your pull request - it should turn green.

Thanks,
Thomas

rattacresh:
Here is the table structure:
```
mysql> DESCRIBE Servers;
+-----------------------+-----------------------------------+------+-----+---------+----------------+
| Field                 | Type                              | Null | Key | Default | Extra          |
+-----------------------+-----------------------------------+------+-----+---------+----------------+
| id                    | int(11)                           | NO   | PRI | NULL    | auto_increment |
| contract_id           | int(11)                           | NO   | MUL | NULL    |                |
| server_type           | enum('dedicated','virtual')       | NO   |     | NULL    |                |
| hostname              | varchar(255)                      | NO   | UNI | NULL    |                |
| ip_v4                 | varchar(15)                       | YES  |     | NULL    |                |
| ip_v6                 | varchar(512)                      | YES  |     | NULL    |                |
| os_type               | enum('Linux','Windows','FreeBSD') | NO   |     | NULL    |                |
| monitoring_node       | varchar(512)                      | NO   |     | NULL    |                |
| monitoring_template   | set('Webserver','Mailserver')     | NO   |     | NULL    |                |
| datacenterlocation_id | int(11)                           | NO   |     | NULL    |                |
| bemerkung             | text                              | YES  |     | NULL    |                |
| wiki_url              | varchar(500)                      | YES  |     | NULL    |                |
+-----------------------+-----------------------------------+------+-----+---------+----------------+
12 rows in set (0.00 sec)
```

and here the query:

```
mysql> SELECT hostname AS object_name, 'object' AS object_type, ip_v4 AS address, ip_v6 AS address6, monitoring_node AS zone, IF (monitoring_template = '','[]',CONCAT('["', REPLACE(monitoring_template, ',', '","'), '"]')) AS import FROM Servers WHERE monitoring_template <> '';
+---------------+-------------+-------------+----------+--------------+----------------------------+
| object_name   | object_type | address     | address6 | zone         | import                     |
+---------------+-------------+-------------+----------+--------------+----------------------------+
| dummyServer_2 | object      | 77.244.96.1 | NULL     | dummyNode_12 | ["Webserver","Mailserver"] |
+---------------+-------------+-------------+----------+--------------+----------------------------+
1 row in set (0.00 sec)
```

array_merge() should simply append the elements in the order specified by the query. That seems to be the natural choice.

Thomas-Gelf:
Merged to `next`, thank you!

PR #950, opened 2017-05-24
Adds sysvinit script

dgoetz:
This pull request adds sysvinit script and some lines of documentation

### DIFF
--- a/contrib/sysvinit/director-jobs.init
+++ b/contrib/sysvinit/director-jobs.init
@@ -1,0 +1,98 @@
+#!/bin/sh
+#
+# director-jobs runs jobs defined in Icinga Web 2 director
+#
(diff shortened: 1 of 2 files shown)


Thomas-Gelf:
Hi @dgoetz,

just to give you a feedback on why I didn't yet merge this: problem is that the CLI command expects to be allowed to exit at any time, as there is systemd restarting the service. So, the implemented logic is "Job finished. Memory usage 'too high'? Exit.". Guess this will rarely ever happen, as jobs are currently forked. Still, the logic is there. As soon as I merge this init script I'm responsible for keeping it running.

So, that's the main reason I didn't touch it yet.

Cheers,
Thomas

Thomas-Gelf:
For reasons mentioned above I have to reject this, sorry

PR #953, opened 2017-05-26
Order and formatting of User, UserGroup and NotificationTable

tobiasvdk:
Order the tables by object name and format disabled objects as in other tables.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,9 @@ env:
     DIRECTOR_TESTDB_USER="director_test"
 
 before_script:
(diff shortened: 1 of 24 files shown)


Thomas-Gelf:
Hi @tobiasvdk,

something went wrong with this pull request, seems that you rebased the master for your branch (instead of doing it the other way round), that's why far too many commits are showing up here. In the meantime, all table-related code has mostly been completely refactored in the `next` branch, so this fixes wouldn't match anyways. But I double-checked right now, seems that all issues you fixed have been addressed so far.

Thanks,
Thomas

PR #958, opened 2017-06-02
QuickTable: Fix enforced filters for non-FilterAnd filters

lazyfrosch:
We need to use the result of andFilter() as the new Filter.

The filter gets replaced by FilterAnd($oldFilter, $enforcedFilter) here.

Please also cherry-pick in 1.3.x


Thomas-Gelf:
Thanks for catching this one!

PR #971, opened 2017-06-17
uncomment to initial support costum plugins

Moep90:


Thomas-Gelf:
Merged to `next`, thank you.

PR #993, opened 2017-07-10
DbObject: Don't confuse ids and object_names

### LINKED ISSUE #992: Creating objects with object_name = some auto increment id wreaks havoc
## Expected Behavior
Creating a host with a number as object_name should be possible without major side-effects. 

## Current Behavior
When I create a host with an object_name that happens to match the internal auto increment id of some other host, any objects referring to that host (like services) get stolen by the newly created host.

## Possible Solution
Cast $id to int before calling static::getPrefetched($id) in DbObject::loadWithAutoIncId(), thus preventing $id to be searched first in the array of object_names.

1. Create some arbitrary host "testhost" and a service for that host
2. insert a host with an object_name that happens to match director's internal auto increment id of testhost
3. Deploy configuration
4. See what host the services for testhost were assigned to.

## Context

We want to assign the ids from our host database to the object_name, for simple and reliable identification, and put the hostname into the display_name instead. We can't simply add a prefix since that would change ticket ids for agents etc., requiring us to redo all hosts.

## Your Environment

* Director version (System - About): 1.3.1
* Icinga Web 2 version and modules (System - About): 2.4.1, director, doc, fileshipper, monitoring
* Icinga 2 version (`icinga2 --version`): r2.6.3-1
* Operating System and version: Debian GNU/Linux 8
* Webserver, PHP versions: apache2 2.4.10-10+de, PHP 5.6.30-0+deb8u1

rattacresh:
fixes #992

### DIFF
--- a/library/Director/Data/Db/DbObject.php
+++ b/library/Director/Data/Db/DbObject.php
@@ -1011,6 +1011,14 @@ abstract class DbObject
 
     public static function loadWithAutoIncId($id, DbConnection $connection)
     {
+        /* Need to cast to int, otherwise the id will be matched against
+         * object_name, which may wreak havoc if an object has a
+         * object_name matching some id. Note that DbObject::set() and
+         * DbObject::setDbProperties() will convert any property to
+         * string, including ids.
+         */
+        $id = (int) $id;
+
         if ($prefetched = static::getPrefetched($id)) {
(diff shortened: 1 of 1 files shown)


Thomas-Gelf:
Merged to `next`, thank you!

PR #1096, opened 2017-08-22
Implement Icinga Dependency Configuration.

mdetrano:
This is an updated implementation of Icinga Dependency configuration in Director.  

Required director database modifications are in:
schema/mysql-migrations/upgrade_dependencies.sql
schema/mysql-migrations/upgrade_dependencies_2.sql

PostgreSQL support is not present (yet...hopefully someone can help with that).

This update makes use of the "suggest" enabled fields on the forms, which should speed up form loading on large systems (long lists of hosts or services for the related objects are not loaded until the ajax call is made for the field).

Now supports "same host" style setups for Apply Rule Dependencies..the "parent host" for such rules can be left blank and, in that case, the service specified is stored as a simple string, instead of attempting to link a specific service for a host.  Otherwise, a parent service is linked internally to a service record, which is related to a host either directly, via a template, or via an apply rule filter.


Thomas-Gelf:
@mdetrano: first of all, thanks a lot for all your efforts and PATIENCE with this feature request! Now that we finally have fields for auto-suggestion this will no longer be an issue for large environments. So we can fine-tune and merge it.

First things I noticed:
* there are some whitespace errors, it would be great if you could fix them - Travis gives related hints
* please re-check required fields, for example I guess "parent" should be required when creating an apply rule
* it breaks symmetry on the new Dashboard :laughing: ... but don't worry, we'll find a solution there
* `parent_service_s` is not a nice name ;)

But apart from this: absolutely great work, thanks a lot! I'll be offline most of this day, but I should be able to look deeper into this pull request later on.

Thanks,
Thomas

cflinspach:
> it breaks symmetry on the new Dashboard ... but don't worry, we'll find a solution there

I'm not sure if it's relevant but it looks like the module pages stick in one column instead of expanding. It looks like this in Brave and Chrome for us.

screenshot from 2017-08-24 08-45-48
screenshot from 2017-08-24 08-49-47
screenshot from 2017-08-24 08-51-35

Thomas-Gelf:
@cflinspach: that's how it will look like in v1.4.0. Using only half of the page is intentional here, as it makes it easier for users to understand where they came from. So this is not related to this pull request.

ktella:
Thomas,

Are you planning to include this in 1.4.0 release ?
I would request to do so.

Thomas-Gelf:
@ktella: well, if you are requesting so... why not :laughing:

Thomas-Gelf:
@mdetrano I merged the pull request into a new branch called `feature/dependencies-1096` and did some cleanup. Mostly spaces, icon, naming - nothing special. In case you need to provide more related commits please use that one as a base for your pull request.

Cheers,
Thomas

vita2:
@mdetrano i've just tested this branch and it works very well! thank you so much for the feature implementation of dependencies 👍 😄  

would it be hard to add the "apply for"-part? so we could configure a dependency apply rule based on custom host vars, like:
```
apply Dependency for (parent in host.vars.parent_hosts) to Host {
  parent_host_name = parent
}
```

Thomas-Gelf:
@mdetrano: as already mentioned, this made it into master :) Thanks again!

Issue #876, opened 2017-03-28
Sync filters have no (or wrong) effect when involved column isn't used otherwise


Thomas-Gelf:
So while setting `address` to `${address}` if `address=127.*` works fine, setting `address`to `127.0.0.1` if `address=127.*` would fail in case `${address}` isn't used elsewhere. Reason: only "used" properties are fetched for Sync execution

lazyfrosch:
See PR #1131

PR #1131, opened 2017-08-29
Sync: Respect the fields used in filter_expression of the SyncRule

### LINKED ISSUE #1130: Filter expression on SyncRule only works when field is used in column filter or properties
The fix in #876 doesn't fix all issues regarding that.

Here is a crude yaml example:

```yaml
rules:
- name: test
  filter_expression: field=*
  properties:
  - source: testdata
    source: ip
    destination: address
```

#876 only fixed that filters on individual properties are taken into account, but the columns used in filter_expression on the rule itself are not taken into account.

/cc @mkayontour @tobiasvdk

### LINKED ISSUE #1141: ImportSource with imports can not be deleted on PostgreSQL
You can not delete ImportSources that had an Import on PostgreSQL.

## Expected Behavior

ImportSource should cascade to deleting all ImportRun in the database.

Works on MySQL.

## Current Behavior

Breaks because of foreign key restriction.

```
SQLSTATE[23503]: Foreign key violation: 7 ERROR: update or delete on table "import_source" violates foreign key constraint "import_run_source" on table "import_run"
DETAIL: Key (id)=(2) is still referenced from table "import_run"., query was: DELETE FROM import_source WHERE (id = '2')
```

## Possible Solution

Update the foreign key constraint in a similar way than MySQL.

## Your Environment

* Director version (System - About): master - 59eda8f9c41f4686a11adc695290700ef54f5c57
* Icinga Web 2 version and modules (System - About): 2.4.1

lazyfrosch:
So that you can filter by fields not used in properties, or in property filters.

fixes #1130
refs #876 

Also fixes #1141 for tests

### DIFF
--- a/library/Director/Import/Sync.php
+++ b/library/Director/Import/Sync.php
@@ -283,6 +283,12 @@ class Sync
                     $filterColumns[$column] = $column;
                 }
             }
+            if (($ruleFilter = $this->rule->filter()) !== null) {
+                foreach ($ruleFilter->listFilteredColumns() as $column) {
+                    $filterColumns[$column] = $column;
+                }
+            }
+
             if (! empty($filterColumns)) {
                 foreach (SyncUtils::getRootVariables($filterColumns) as $column) {
                     $usedColumns[$column] = $column;
--- a/library/Director/Objects/SyncRule.php
+++ b/library/Director/Objects/SyncRule.php
@@ -203,7 +203,10 @@ class SyncRule extends DbObject
         return $this->sync;
     }
 
-    protected function filter()
+    /**
+     * @return Filter
+     */
+    public function filter()
     {
         if ($this->filter === null) {
             $this->filter = Filter::fromQueryString($this->get('filter_expression'));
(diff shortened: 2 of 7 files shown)


lazyfrosch:
**NOTE:** Currently untested, lacking a proper test environment.

Please test and give feedback @mkayontour

lazyfrosch:
@mkayontour please also test if this works with your multi-level column `guest.ipAddress`!

Thomas-Gelf:
@lazyfrosch: thank you, this was indeed missing. I also didn't try it out, but it looks good to me.

Thomas-Gelf:
Ping @mkayontour

lazyfrosch:
The behavior intended to fix should work.

@Thomas-Gelf I've added a test for Sync, please review. It tests for basic Import/Sync behavior and filters. Still extendable :laughing:

lazyfrosch:
Fixed a small mistake for PostgreSQL

lazyfrosch:
Seems like I have to take a closer look on testing with PostgreSQL

lazyfrosch:
PostgreSQL tests fail because of #1141

lazyfrosch:
@Thomas-Gelf ready 4 merge / review / refactor :tongue:

Thomas-Gelf:
@lazyfrosch: well done, thank you!

PR #1158, opened 2017-09-06
Update 70-REST-API.md

ncitop:
Variable "object_name" is needed to register a new host via the self-service API.

### DIFF
--- a/doc/70-REST-API.md
+++ b/doc/70-REST-API.md
@@ -540,6 +540,7 @@ A new node will `POST` to `self-service/register-host`, with two parameters in
 the URL:
 
 * `name`: it's desired object name, usually the FQDN
+* `object_name`: Name that will be displayed
 * `key`: a valid Host Template API key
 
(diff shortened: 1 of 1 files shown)


kajla:
Thanks!

hrak:
This is not true based on my experience. A POST to director/self-service/register-host with just name and key works fine with director 1.5.1.

raviks789:
Closing this as stale. I checked the current code against the claim here: object_name is only required when the name URL parameter is omitted from the request (`IcingaHostSelfServiceForm::setup()` drops the `object_name` field entirely
  once `setHostName()` is called). Since the documented flow passes both name and key as URL params, `object_name` isn't needed there, which matches what @hrak reported back in 2018. So the change as written would overstate the requirement rather than clarify it.

If the docs should call out the alternate body-only object_name path as an option, happy to see a fresh PR against current master that describes it accurately.

Issue #1157, opened 2017-09-06
Separate permissions for host and service templates


matthenning:
## Expected Behavior

Enable users to manage their own templates by granting them permission based on e.g. custom vars.

## Current Behavior

Only single services or hosts are covered by the director/services, director/hosts permissions.
Templates are not editable.

## Possible Solution

Add a separate permission, e.g. director/service_templates and director/host_templates and add a filter attribute like the existing director/filter/hostgroup

Thomas-Gelf:
I could implement this based on service template names (forcing them to match a specific pattern) very soon if you want. This will allow you to grant `director/service/template` and to restrict `director/service/template/filter-by-name = "* by Tom"`. This would allow modify (and create) access to all templates whose name is ends with " by Tom".

Please note that documentation will warn and teach you to grant this permission to trusted users only. Someone able to modify Templates can easily execute any command he wants on your systems, you can only restrict him providing a very limited set of CheckCommands. Those shipped with the ipl would already be problematic. That's why templates used to be an admin-only feature. Also, one could try to DoS your systems by changing the check_interval to 0.02s, applying that service multiple times to all your hosts.

matthenning:
Sounds great.

We will only allow certain admins access to the Director with specific training to minify the risk.

Thomas-Gelf:
According to the code, such restrictions should already be in place. They are not shown/exported to the permissions/roles admin UI, and unfortunately it's now too late to seriously test this for v1.7.0. Skipping to the next version

EmRowlands:
Will this feature be coming to a future release? This would be really useful

PR #1166, opened 2017-09-08
Separate permissions for host and service templates

matthenning:
Implements #1157 

Somehow the director/host_templates permission doesn't show up in my test case and I can't seem to figure out why. director/service_templates works fine.

I hope I caught every occurence of the director/admin permissions. My test seem fine.


aha2140:
+1 for this feature. What are the plans for this to be merged in?

EmRowlands:
Is this still planning on being merged? It looks like it will need a rebase, but it's a feature that I'd quite like to have.

Issue #1209, opened 2017-10-04
Allow sync for object type dependency


dgoetz:
## Expected Behavior
Allow to sync dependency with all its attributes and assign where filters.

## Current Behavior
Sync rules do not allow to select the object type dependency.

## Context
In the import source are a couple of routers which are the parent host for other hosts in the same location. Creating and maintaining them automatically instead of manually would reduce the workload.

## Your Environment
* Director version (System - About): Git Master 6980544
* Icinga Web 2 version and modules (System - About): 2.4.1
* Icinga 2 version (icinga2 --version): 2.7.1
* Operating System and version: RHEL 7
* Webserver, PHP versions: Apache 2.4, PHP 5.4.45

darkounet:
+1 👍

bahamut45:
+1 👍

vita2:
sorry for the ref spam. there is something gone wrong :confused:

PR #1212, opened 2017-10-04
Sync rule: allow dependencies

vita2:
with this PR it is allowed to sync dependencies.

the part for PostgreSQL is missing because i haven't any experience with it. i could not test the changes.

refs #1209


vita2:
i cannot merge this PR after applying PR #1211. there are merge conflicts with `schema/mysql.sql` and `application/forms/SyncRuleForm.php`. 

how could i fix this? thought it would be better to split the changes regarding sync rules in two separate pull requests  :disappointed:

PR #1255, opened 2017-10-23
Change main menu name

rockyluke:
Rename "Icinga Director" to "Director" like all other modules (no "Icinga" in front of the name).

### DIFF
--- a/configuration.php
+++ b/configuration.php
@@ -75,7 +75,7 @@ $this->provideConfigTab('config', array(
 ));
 
 $section = $this->menuSection(
-    N_('Icinga Director')
+    N_('Director')
 )->setUrl('director')->setPriority(60)->setIcon(
     'cubes'
 )->setRenderer(array(


walterheck:
@Thomas-Gelf this is a fairly trivial PR, can it be merged please?

Thomas-Gelf:
@walterheck: I'm not sure about this one, as there are too many tools called `Director` out there. Also, the compound variant looks better next to other longer names, at least as soon as a couple of non-standard modules have been activated.

I'd prefer to reject this one, but I'll go and ask for other opinions on this.

rockyluke:
Thanks !

friesoft:
It makes sense to call the product "Icinga Director" (e.g. for easing searching) - but I agree with the above commenters that "Director" would look better in the main menu.

rockyluke:
That's my point :)

Thomas-Gelf:
Sorry guys, I have to refuse this. We had a quick related discussion, for now it will continue to read "Icinga Director". Depending on which modules you're running one or the other fits better. My current local setup (80+ modules, 40+ of them currently enabled) looks better with the long name, as it sits between "Business Processes" and "VMware vSphere DB". Your menu might of course look different.

Btw, there is also a "vCloud Director", who might sooner or later be worth a dedicated module. Time ago we evaluated some Bareos/Bacula integration, they also have a Director. So, just "Director" might become confusing. We also discussed renaming it to "Configuration", but that's too generic. "Icinga Configuration" then once again carries Icinga.

Once we release the upcoming IcingaDB, there could be a deeper integration of our monitoring visualization and it's configuration. This could eventually result in less dedicated menu entries. Parts of the Director could be isolated and provided as building blocks to other modules in a more generic way. In addition, we see more and more nice modules being released, this might sooner or later push us towards adding an additional navigation layer - we'll see.

Long story short: we want to find a better naming, but just calling it "Director" doesn't feel being the better option. So for now it will remain as is.

Anyways, thanks for your effort!

friesoft:
Thanks for the detailed explanation! :) Looking forward to the bright future of the Icinga Stack :)

rockyluke:
Thanks for the detailed explanation !!

walterheck:
Just to be clear: I'm totally on board with this either way, I just activated the discussion because it was stale for a while. This project has 300+ issues and 20+ PR's and that seems unsustainable :)

PR #1260, opened 2017-10-25
Working with Powershell Scripts

gitlo88:
I will create a howto for adding powershell scripts using icinga2 director.


gitlo88:
First Version. On the next version i will add pictures and optimizations.

PR #1285, opened 2017-11-15
ObjectController: Don't let DeploymentLink break the interface on errors

lazyfrosch:
This can happen whenever the deployment related config of Zones,
Endpoints or ApiUsers is incomplete or missing.

This should not break forms needed to fix the problem.


lazyfrosch:
@Thomas-Gelf Please have a look

Issue #1320, opened 2017-12-06
Sync rule for group membership can't handle comma separated list


bl4kd3th:
## Expected Behavior
When providing comma separated list of groups, it should add the list of groups to the host object.

## Current Behavior
Throws an error stating your entered text is not the name of a group.
![image](https://user-images.githubusercontent.com/1704681/33686594-05a13f1a-da9b-11e7-922b-59d188969a25.png)

## Possible Solution
Add explode() on comma. Will send PR for it.

## Steps to Reproduce (for bugs)
1. Add import source
2. Add sync rule. 
3. Add a property similar to this one: 
![image](https://user-images.githubusercontent.com/1704681/33686695-599b71b2-da9b-11e7-9a99-b0c543b35b7b.png)

4. Click "Store", then click the "Check for changes" button on the sync rule.
5. Wait a moment or refresh the pane to see the error similar to:
![image](https://user-images.githubusercontent.com/1704681/33686594-05a13f1a-da9b-11e7-922b-59d188969a25.png)

## Context
I wanted to bulk-update a list of hosts by adding multiple hostgroups to each one. If I add multiple "Group membership" properties, they just overwrite each other and you end up with only the last one "Group membership" property.

## Your Environment

* Director version (System - About): 1.4.2 (but really I have clone of master branch from 2017-11-21)
* Icinga Web 2 version and modules (System - About): 2.5.0
* Icinga 2 version (`icinga2 --version`): r2.8.0-1
* Operating System and version: CentOS Linux release 7.4.1708 (Core)
* Webserver, PHP versions: httpd 2.4.6-67.el7.centos.6, php 7.1.8-1.el7

Thomas-Gelf:
I rejected your pull request, please use the split (or regex based split) Import Source property modifier to accomplish what you need.

bl4kd3th:
@Thomas-Gelf ah that is a much better way to handle it. Thank you!

Thomas-Gelf:
You're welcome!

PR #1321, opened 2017-12-06
add comma separated list handling for groups membership

bl4kd3th:
Reference issue: #1320


Thomas-Gelf:
I have to reject this, a comma is a valid character in an Icinga 2 group name.

Thomas-Gelf:
I have to reject this, a comma is a valid character in an Icinga 2 group name.

PR #1346, opened 2017-12-23
Support for per zone hostgroups.

nunofernandes:
In my setup, I have a icinga master node and several satellites that belong to different companies.
In the current implementation, all hostgroups are replicated to all satellites and our costumers know about each others hostgroups.

This patch creates a zone_id for the hostgroup and creates the hostgroups' configuration file in their specific rendering zone.

We require a mysql schema change:
```
ALTER TABLE icinga_hostgroup ADD zone_id INT(10) UNSIGNED DEFAULT NULL;

ALTER TABLE icinga_hostgroup ADD CONSTRAINT icinga_hostgroup_zone FOREIGN KEY zone (zone_id) REFERENCES icinga_zone (id) ON DELETE RESTRICT ON UPDATE CASCADE;
```

Do you think that this makes sense or can we do it in a different way? If you agree, please merge this. If you don't, please let me know how can i improve the patch so that is gets merged.

Nuno Fernandes

### DIFF
--- a/application/forms/IcingaHostGroupForm.php
+++ b/application/forms/IcingaHostGroupForm.php
@@ -18,8 +18,26 @@ class IcingaHostGroupForm extends DirectorObjectForm
 
         $this->addGroupDisplayNameElement()
              ->addAssignmentElements()
+             ->addZoneElements()
              ->setButtons();
     }
+
+    protected function addZoneElements()
+    {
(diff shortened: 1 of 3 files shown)


nunofernandes:
Please ignore the pull request as, by mistake there are 2 commits and the second one isn't related to the subject of the pull request.

PR #1349, opened 2017-12-28
Add support for per zone hostgroups

nunofernandes:
In my setup, I have a icinga master node and several satellites that belong to different companies.
In the current implementation, all hostgroups are replicated to all satellites and our costumers know about each others hostgroups.

This patch creates a zone_id for the hostgroup and creates the hostgroups' configuration file in their specific rendering zone.

We require a sql schema change (already updated in the sql files):
```
ALTER TABLE icinga_hostgroup ADD zone_id INT(10) UNSIGNED DEFAULT NULL;

ALTER TABLE icinga_hostgroup ADD CONSTRAINT icinga_hostgroup_zone FOREIGN KEY zone (zone_id) REFERENCES icinga_zone (id) ON DELETE RESTRICT ON UPDATE CASCADE;
```

Do you think that this makes sense or can we do it in a different way? If you agree, please merge this. If you don't, please let me know how can i improve the patch so that is gets merged.

### DIFF
--- a/application/forms/IcingaHostGroupForm.php
+++ b/application/forms/IcingaHostGroupForm.php
@@ -18,8 +18,26 @@ class IcingaHostGroupForm extends DirectorObjectForm
 
         $this->addGroupDisplayNameElement()
              ->addAssignmentElements()
+             ->addZoneElements()
              ->setButtons();
     }
+
+    protected function addZoneElements()
+    {
+        $this->addZoneElement();
+        $this->addDisplayGroup(array('zone_id'), 'clustering', array(
+            'decorators' => array(
+                'FormElements',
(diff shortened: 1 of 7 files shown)


nunofernandes:
Aparently the build error is related to previous code

```
1) Tests\Icinga\Module\Director\IcingaConfig\StateFilterSetTest::testIsEmptyForAnUnstoredUser
Icinga\Exception\IcingaException: Migration 0 failed (SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'director_activity_log' already exists) while running CREATE TABLE director_activity_log (
```

Maybe related to the fact that I added some files to the {pg,my}sql-migration folders. Need help in debugging this build issue.

PR #1351, opened 2017-12-28
Bugfix/timeperiodrangeprint

nunofernandes:
When i have a timeperiod without defined ranges (they are imported from the template), the director code generates a configuration like:

```
object TimePeriod "timeperiod-10x5" {
    import "legacy-timeperiod"
    import "timeperiod-template-10x5"

    display_name = "timeperiod-10x5"
    ranges = {
    }
}
```

It generates an empty ranges. This patch checks if the ranges array is empty and if so, i doesn't generate any timerange.
After patch we get:

```
object TimePeriod "timeperiod-10x5" {
    import "legacy-timeperiod"
    import "timeperiod-template-10x5"

    display_name = "timeperiod-10x5"
}
```


Thomas-Gelf:
Depending on your point of view this absolutely makes sense. Just, changing it right now would change behavior for existing systems in an unexpected way. I'll leave this pull request open, eventually we can find a better solution.

Thomas-Gelf:
I'll close this pull request for now. Still, I'd love to see an extra flag allowing one to achieve the behavior you asked for.

PR #1353, opened 2017-12-29
fix a copy&paste error in health check documentation

widhalmt:
There was an copy&paste error in the health check documentation

I added another example with the short form of the check just in case we need the short form for the ITL integration.


Thomas-Gelf:
Thanks!

PR #1354, opened 2017-12-29
add Zone option to Dependency Form

mdetrano:
I'd like to propose this small change to "Dependencies", which allows users to force a Dependency into a specific zone.  This is useful when someone wants to create a dependency between an object in one zone to an object in a parent zone.  If the Dependency apply rule is defined in the global zone (current behaviour), an Endpoint in the child zone may try to apply the dependency to its object, but will throw an error because it does not know about the parent object in another zone.  Forcing the Dependency only to deploy to the parent zone fixes that problem.


Thomas-Gelf:
Makes sense, thank you!

PR #1397, opened 2018-02-20
ServiceActions: Show inspect link on all services

lazyfrosch:
Not only services of hosts managed by Director. This aligns behavior
with HostActions itself.


lazyfrosch:
@Thomas-Gelf small fix, please review

Thomas-Gelf:
Makes sense, thank you

Issue #944, opened 2017-05-22
Cannot add Timeperiod to User


hermann1514:
In Director is it not possible to add a specific timeperiod to a User in the notification Menu's.
Bug?

Thomas-Gelf:
Hi @hermann1514,

thanks for reporting this, it's a bug. Most people define periods in their Notification Apply rules, guess that's why nobody complained yet :laughing:

Will be fixed, in the meantime you could use the CLI as a workaround:

```
icinga@web1:~# icingacli director user set tom --period 24x7
User 'tom' has been modified
icinga@web1:~# icingacli director user show tom              
object User "tom" {
    import "admin user"

    display_name = "Thomas Gelf"
    period = "24x7"
}
```

If you set periods for a few User Templates via CLI, you could then easily use them in the GUI. Or just wait for the fix ;-)

Cheers,
Thomas

hermann1514:
Great. It works now.
Thx.
Hermann

hermann1514:
Hi, after an update from Director it seems so that the Commandline icingacli director user set not working. I can remove a period but when i add an List the User was not modified.

hermann1514:
In Version 1.3.2 it's working

Not in the Master Version

hermann1514:
Hi, 

with the new Version from Director 1.4 i have the same Problem.
I can remove a period from  a user with

iciingacli director set user xxxxxxx --period ""

but with 

iciingacli director set user xxxxxxx --period "Periodname"
the User object was not modified and the new Period is not set to the User.

bargru:
@hermann1514 the bug is fixed for me with Director 1.4.3

mwaldmueller:
+1

PR #1398, opened 2018-02-20
Allow to set a TimePeriod for an User via GUI

bargru:
With the CLI, a time period could already be assigned to a user. With this patch it is also possible via the GUI. The added code is copied from the IcingaNotificationForm.

Relates to this issue #944

grafik


bargru:
It's now possible in master. So i close this pr.

PR #1413, opened 2018-02-28
IcingaConfig: Remove library import from generated config

lazyfrosch:
This hasn't been used since Director doesn't set execution methods for commands anymore.

Code cleanup deluxe...


lazyfrosch:
On further investigation the addition here is no longer needed, we can remove those lines.

lazyfrosch:
@Thomas-Gelf I guess we can merge this now?

Thomas-Gelf:
Merged, thank you

PR #1414, opened 2018-02-28
In service template it's possible to add a command template in the check_command field

ValeDaRold:
In Add Service Template form you can add a 'Command Template' in 'Check command' field: the field itself shows up both commands and command templates.
image
Once you've created a service based on a service template with a command template as 'Check command' property, the deploy fails, throwing this error:
`critical/config: Error: Validation failed for object 'Standard Linux Server!Service with Command Template' of type 'Service'; Attribute 'check_command': Object 'Command Template' of type 'CheckCommand' does not exist.`
image
For this reason I think that could be better to filter Command in advance and show up in 'Check command' field only real 'Command', excluding 'Command Templates' `"object_type != 'template'"`.

### DIFF
--- a/application/controllers/SuggestController.php
+++ b/application/controllers/SuggestController.php
@@ -220,7 +220,7 @@ class SuggestController extends ActionController
 
     protected function suggestServiceFilterColumns()
     {
-        return $this->getFilterColumns('host.', [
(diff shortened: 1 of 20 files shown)


Thomas-Gelf:
Thanks for this!

PR #1416, opened 2018-03-01
IcingaArguments: allow to manage functions through API

capull0:
We solved our bottom-up configuration via the director API and a python script.
With this patch it is also possible to handle functions through the API

Works for us, but don't know if it breaks something else.
I don't know the reason for:
$attrs['argument_value'] = '/* Unable to fetch function body through API */';

### DIFF
--- a/library/Director/Objects/IcingaArguments.php
+++ b/library/Director/Objects/IcingaArguments.php
@@ -140,13 +140,21 @@ class IcingaArguments implements Iterator, Countable, IcingaConfigRenderer
             }
             if (property_exists($value, 'type')) {
                 if ($value->type === 'Function') {
-                    $attrs['argument_value'] = '/* Unable to fetch function body through API */';
+                    if (property_exists($value, 'body')) {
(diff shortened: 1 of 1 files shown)


mdetrano:
Hello,

I am interested in this pull request and have tested it independently.

It think in the patch, around line 153, this line:

                         if (property_exists($value, 'body')) { 


should be

                        if (property_exists($value->value, 'body')) {


also,
in the same function (mungeCommandArgument) , if this is added to the $map variable:
            'set_if_format'      => 'set_if_format',
Then the set_if parameters that are specified with Icinga DSL will be imported correctly (otherwise this values is left as "string"...which it isn't).

Combining the changes in this pull request with the couple of fixes stated here, I can get the full cycle to work in the API...  GET json data, then POST the same data to create a working command object just as it was originally.

This also helps greatly when using the "restore object" feature in Director's activity log to recover a deleted command (the DSL stuff is restored as well).

Thomas-Gelf:
Thanks for this pull request, this should have been implemented with 6c00d17.

PR #1417, opened 2018-03-01
Update Icinga2Agent.bash

michaeleino:
correcting object ApiListener "api" ca_path

to be the same as created /icinga2/pki/trusted-master.crt instead of /icinga2/pki/ca.crt

makes icinga2 service not able to start due to missing file.


michaeleino:
sorry, it was confusing, think it is fine!

Issue #1422, opened 2018-03-08
ImportRunBasedPurgeStrategy performance issue


lazyfrosch:
*Fair warning:* In the depths of director.

While working on bigger sync jobs I noticed that Sync runs can take a long of time without making sense at all.

I tracked this down to `ImportRunBasedPurgeStrategy` so far, culprit seems to be this query:

https://github.com/Icinga/icingaweb2-module-director/blob/e9a570e96d77f7f083ba5f366b94ceb09582975f/library/Director/Import/PurgeStrategy/ImportRunBasedPurgeStrategy.php#L55-L67

With my test data I want to import around 6k hosts, with only some hosts present currently.

The execution of this query takes around 45-55 seconds!

With some benchmarking, these are the raw numbers for 2 single, and the combined query.

```
Purge: Begin finding keys	824.336	2.211	22.60 KiB	13.28 MiB
Query A: 7032	835.341	11.005	3.32 MiB	16.60 MiB
Query B: 7032	847.241	11.900	3.32 MiB	19.92 MiB
Query Sub-Select: 0	51325.518	50478.277	496.00 B	19.92 MiB
```

I have no idea what is going on with the database server: `Ver 15.1 Distrib 10.2.11-MariaDB`

lazyfrosch:
Crude debug snippet before the `fetchCol()` call:

```php
        echo $selectA;
        $a = $db->fetchAll($selectA);
        Benchmark::measure('Query A: ' . count($a));

        echo $selectB;
        $b = $db->fetchAll($selectB);
        Benchmark::measure('Query B: '. count($b));

        echo $query;
        $all = $db->fetchAll($query);
        Benchmark::measure('Query Sub-Select: ' . count($all));

        echo Benchmark::dump();

        echo Util::wantHtml($db->fetchAll('EXPLAIN ' . $query))->render();

        exit;
```

Thomas-Gelf:
Can you show me the `EXPLAIN` output? 6000 objects shoultn'd hurt at all. And please also:

```sql
SELECT COUNT(*) FROM imported_property;
SELECT COUNT(*) FROM imported_row_property;
SELECT COUNT(*) FROM imported_rowset_row;
SELECT COUNT(*) FROM imported_property;
```

Thomas-Gelf:
Oh, and your `innodb_buffer_pool_size` would be interesting.

lazyfrosch:
[...]

PR #1426, opened 2018-03-11
Improve performance of sync

lazyfrosch:
This addresses multiple problems in Sync and GroupMembershipResolver.

- [x] Add Benchmark at more code locations
- [x] Sync: Prefetch related objects
- [x] GroupMembershipResolver: Parser Filter before iteration
- [x] ~~ImportRunBasedPurgeStrategy performance issue #1422 (discussed there)~~
- [x] ObjectApplyMatches: Prepare filters only once

### DIFF
--- a/application/controllers/SyncruleController.php
+++ b/application/controllers/SyncruleController.php
@@ -26,6 +26,9 @@ class SyncruleController extends ActionController
         $ruleName = $rule->get('rule_name');
         $this->addTitle($this->translate('Sync rule: %s'), $ruleName);
 
+        $checkForm = SyncCheckForm::load()->setSyncRule($rule)->handleRequest();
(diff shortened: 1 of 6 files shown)


lazyfrosch:
The membership resolver is still not very fast, still digging deeper:

Description | Off (ms) | Dur (ms) | Mem (diff) | Mem (total)
-- | -- | -- | -- | --
Hostgroup apply recalculated - objects=3433 groups=1837 min=5 max=77 avg=6 (in ms) | 27858.643 | 21135.931 | 25.41 MiB | 84.76 MiB

Thomas-Gelf:
Well, one note - just for the records: not redirecting after form submission is technically not correct, but we have no other easy way to keep state right now. We could use the session, but for now - this is fine. Go!

Issue #1325, opened 2017-12-07
ImportSource preview/history has problem rendering certain types


lazyfrosch:
Certain types fail to render in the table.

    String, Html Element or Array of such expected, got "stdClass" (Util.php:106)
    String, Html Element or Array of such expected, got "boolean" (Util.php:106)

```
#0 /usr/share/icingaweb2/modules/director/library/vendor/ipl/Html/Html.php(34): dipl\Html\Util::wantHtml(Object(stdClass))
#1 /usr/share/icingaweb2/modules/director/library/vendor/ipl/Html/BaseElement.php(119): dipl\Html\Html->add(Object(stdClass))
#2 /usr/share/icingaweb2/modules/director/library/vendor/ipl/Html/Html.php(123): dipl\Html\BaseElement->add(Object(stdClass))
#3 /usr/share/icingaweb2/modules/director/library/vendor/ipl/Html/Html.php(113): dipl\Html\Html->addContent(Object(stdClass))
#4 /usr/share/icingaweb2/modules/director/library/vendor/ipl/Html/Table.php(161): dipl\Html\Html->setContent(Object(stdClass))
#5 /usr/share/icingaweb2/modules/director/library/vendor/ipl/Web/Table/QueryBasedTable.php(168): dipl\Html\Table->renderRow(Object(stdClass))
#6 /usr/share/icingaweb2/modules/director/library/vendor/ipl/Web/Table/QueryBasedTable.php(116): dipl\Web\Table\QueryBasedTable->fetchRows()
#7 /usr/share/icingaweb2/modules/director/library/vendor/ipl/Html/BaseElement.php(133): dipl\Web\Table\QueryBasedTable->renderContent()
#8 /usr/share/icingaweb2/modules/director/library/vendor/ipl/Html/Html.php(171): dipl\Html\BaseElement->render()
#9 /usr/share/icingaweb2/modules/director/library/vendor/ipl/Html/BaseElement.php(105): dipl\Html\Html->render()
#10 /usr/share/icingaweb2/modules/director/library/vendor/ipl/Html/BaseElement.php(133): dipl\Html\BaseElement->renderContent()
#11 /usr/share/icingaweb2/modules/director/library/vendor/ipl/Html/Html.php(259): dipl\Html\BaseElement->render()
#12 /usr/share/icingaweb2/modules/director/library/vendor/ipl/Zf1/SimpleViewRenderer.php(47): dipl\Html\Html->__toString()
[...]

PR #1427, opened 2018-03-11
ImportedrowsTable: Render complex data types in pre tag

lazyfrosch:
Just like ImportsourceHookTable.

refs #1325


lazyfrosch:
I guess this is a better fix than https://github.com/Icinga/icingaweb2-module-director/commit/50116a1e4e68d3c8cffe5c6980f0a78d1c94febf 

But maybe we should move the `renderRow()` function to a parent class under `SimpleQueryBasedTable`

@Thomas-Gelf ?

PR #1429, opened 2018-03-12
Fix settings name in PropertyModifierMakeBoolean.

pprkut:
This bug is also present in support/1.3 and support/1.4, but I'm not sure what the merging policy is for those. I can rebase to either of those if needed, or file separate pull requests.


Thomas-Gelf:
Thank you!

PR #1431, opened 2018-03-16
Add conditional replacement property modifier

offsides:
This is an additional property modifier that allows you to replace a target property with the value of the source property only if the requisite condition is met.  Currently implemented conditions are:

Replace only if the target property doesn't exist
Replace only if the target property doesn't exist or is null
Replace only if the target property doesn't exist or is an empty string
Replace only if the target property exists and is an empty string (sets target to null if it doesn't exist)
Replace only if the target property exists and is not a true value (sets target to null if it doesn't exist)
Replace only if the target property exists and is a true value (sets target to null if it doesn't exist)

### DIFF
--- a/library/Director/PropertyModifier/PropertyModifierCondReplace.php
+++ b/library/Director/PropertyModifier/PropertyModifierCondReplace.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace Icinga\Module\Director\PropertyModifier;
+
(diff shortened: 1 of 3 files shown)


offsides:
I messed up my github commit due to inexperience, so I'm closing this one and will resubmit after fixing both that and the issues that the test found.

Issue #619, opened 2016-11-23
[dev.icinga.com #13283] Missing Groups fiel for services through template


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/13283

**Created by birkir on 2016-11-23 01:54:03 +00:00**

Assignee: _(none)_
Status: _Assigned_ 
Target Version: _(none)_
Last Update: _2016-12-07 00:02:41 +00:00 (in Redmine)_

---

When creating a service for a host template, it is not possible to set a service-group for that service.

Create new host template -&gt; Add service -&gt; no Groups field.

dgoetz:
I have the same problem with services applied as a service set. The use case is to have the same checks like basic http check (and very specific ones) used in different service sets for different application server. So in this case it makes no sense to define group membership on the service template and apply form already allows to set the groups, so it make sense to have this option also on service sets. If you agree it would be a one line change if I am correct (and a second similar one for the original request).

```
diff --git a/application/forms/IcingaServiceForm.php b/application/forms/IcingaServiceForm.php
index 6358345..dc06210 100644
--- a/application/forms/IcingaServiceForm.php
+++ b/application/forms/IcingaServiceForm.php
@@ -310,6 +310,7 @@ class IcingaServiceForm extends DirectorObjectForm
 
         $this->addNameElement()
              ->addDisabledElement()
+             ->addGroupsElement()
              ->groupMainProperties();
 
         if ($this->hasPermission('director/admin')) {
```

Another idea would be to have the groups element on the service set form and let services inherit from it but this would require more work including a new database table. Also it will make configuration easier for some but will perhaps not fit everyone's use case, so I would opt for the first one. 

If you agree I can create a pull request for this.

dgoetz:
The pull request adds the element to add Servicegroups in Service sets and at Host templates.

The logic with inherited and applied groups like it is available for Hostgroups is still missing as it would require a similar resolver and instead of duplicating the code it could be perhaps moved to be more general. But this is to much for my coding skills. ;-)

Thomas-Gelf:
Well... it's merged, so let's mention it.

PR #1433, opened 2018-03-19
Add Servicegroups directly to Services

dgoetz:
As discussed in #619

### DIFF
--- a/application/forms/IcingaServiceForm.php
+++ b/application/forms/IcingaServiceForm.php
@@ -267,6 +267,7 @@ class IcingaServiceForm extends DirectorObjectForm
         $this->addNameElement()
              ->addChoices('service')
              ->addDisabledElement()
+             ->addGroupsElement()
              ->groupMainProperties()
              ->addCheckCommandElements()
              ->addExtraInfoElements()
@@ -310,6 +311,7 @@ class IcingaServiceForm extends DirectorObjectForm
 
         $this->addNameElement()
              ->addDisabledElement()
+             ->addGroupsElement()
              ->groupMainProperties();
 
         if ($this->hasPermission('director/admin')) {


Damoun:
Hi,

What's the state of this PR ? I just experienced the need for this feature :D

Thomas-Gelf:
Reason for not merging this was:

* it should not be up to the end user to decide which group a Service belongs to
* in nearly all the cases this should be pretty clear once you define your Service template
* in case it isn't, you're probably doing templates wrong
* examples might be legacy ugliness like one generic nrpe template while having all the details in your Services - like in the linked discussion
* service groups might become a security feature, like host groups. Once restricted, you're no longer allowed to touch them

I delayed as I wasn't yet sure about all of it's implications and honestly didn't invest time to reflect it. It's now in the master, so what - I wasn't strictly against it. Still, I'm very curious to learn about other real-world use-cases for this feature. Is there anything else, apart from the mentioned NRPE-BS?

ping @dgoetz, @lazyfrosch.

lazyfrosch:
In my view I want to support the basic features of Icinga 2.

E.g. whats the point of being able to sync a param, but not being able to set in the UI.

In terms of restrictions for "users" this has to be addressed in a different way...

Thomas-Gelf:
Sync is an Admin-only feature, Templates are Admin-only features, single Services are not. Services should be as lean as possible and are to be seen as an "End User Feature". Mapping every single parameter in every place was never the goal for the Director - and it should never be. Otherwise it should better have been a web-based config file editor.

Looking forward to learning about real-world use-cases for this feature.

Thanks,
Thomas

Thomas-Gelf:
Valid points. But. In my naive view of the world, "Application" is a property - and should be a Custom Variable. Same goes true for "High Priority". Now if someone in addition to those flags wants to also have Service Groups, they can be assigned via Apply Rules based on those properties.

Thoughts?

PR #1434, opened 2018-03-20
Property modifier to conditionally replace another property value

offsides:
This modifier allows you to conditionally replace the value of one property with the value of another.  The conditions currently supported are:

  Replace if target doesn't exist
  Replace if target doesn't exist or is null
  Replace if target doesn't exist or is empty
  Replace if target exists and empty (sets to null if doesn't exist)
  Replace if target exists and not true (sets to null if doesn't exist)
  Replace if target exists and is true (sets to null if doesn't exist)

This makes it easier to manipulate data from multiple columns that may or may not be populated in all rows (e.g., vSphere) and only override values when necessary.

### DIFF
--- a/library/Director/PropertyModifier/PropertyModifierCondReplace.php
+++ b/library/Director/PropertyModifier/PropertyModifierCondReplace.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace Icinga\Module\Director\PropertyModifier;
+
+use Icinga\Module\Director\Hook\PropertyModifierHook;
+use Icinga\Module\Director\Import\SyncUtils;
+use Icinga\Module\Director\Web\Form\QuickForm;
+
+class PropertyModifierCondReplace extends PropertyModifierHook
+{
+    public static function addSettingsFormFields(QuickForm $form)
+    {
+        $form->addElement('select', 'replace_on', array(
+            'label'        => $form->translate('Replacement condition'),
+            'description'  => $form->translate('Under what condition should we replace the target value?'),
+            'multiOptions' => $form->optionalEnum(array(
+                'missing' => $form->translate("Replace if target doesn't exist"),
+                'null'    => $form->translate("Replace if target doesn't exist or is null"),
+                'blank'   => $form->translate("Replace if target doesn't exist or is empty"),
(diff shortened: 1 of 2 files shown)


Thomas-Gelf:
Could you please share some sample data that would allow us to better understand what kind of problem you're trying to solve? At first sight this looks like something that is usually solved at Sync time.

lazyfrosch:
Most of the checks should be converted to `empty()` instead of `== ''` or `if ($var)` or `if (!$var)`

offsides:
Unfortunately, my Icinga2 project has been sidelined for long enough that I don't remember the exact situation I was working with, but IIRC I was attempting to work with vSphere imports to assign a proper address to each VM, and the assignment is conditionally dependent on both the existing value in the raw imported data (i.e., if it has one or not), as well as the result of a DNS lookup on the VM name and possibly another factor.  Basically, there are going to be times where you want to do a replacement of a field value, but only if the import source (or modified source field) doesn't have one in that column to begin with.  The existing (at least at the time, I haven't played with the latest releases) modifiers only work either via regex on an existing value, or by unconditionally replacing the target value.  This provides a larger range of conditions, as well as supporting cases where the field is null or doesn't exist in the first place (e.g., because it was created as the target of a regex condition, but the regex fails on some fields).

Thomas-Gelf:
What you describe is usually easy to solve in the Sync Rule. Define the same property multiple times, from multiple source columns or even different Import Sources. Sort them up or down as required, last non-NULL-value wins.

offsides:
Hmm...  I'll have to investigate this a little further.  However, when I just looked at things I did notice something slightly off.  When adding a sync property you cannot specify the destination column.  You can only do that when you modify the property after adding it.  Not sure if that was changed for the latest release or not (I'll update to that before I do anything else), but that may have been partially why I didn't know that was an option.  Again, it's been a while since I spent any time on this in depth, so I'll have to look at it further to see if your suggestion does what I need or not.  Feel free to put further discussion of this on hold until I can do so and provide further feedback.

Thomas-Gelf:
@offsides: any news on this? Do you still see a need for this feature?

Thomas-Gelf:
Closing as of no feedback. Please do not hesitate to open an updated pull request in case you still need this.

PR #1436, opened 2018-03-20
Add 'is false (or not set)' condition for apply rules

offsides:
Allow rules to apply if a value is false or not set.


offsides:
It appears that Travis CI had an internal problem and it's not the code.  Will it automatically retest or do I need to do something to have it do so?

PR #1448, opened 2018-03-28
Allow creating hooks for custom fields in host/service form

ValeDaRold:
Allow creating non editable fields via hook in Director host/service form by registering them in IcingaWeb  modules.


Thomas-Gelf:
Merged, thank you!

Issue #1446, opened 2018-03-28
Dependency: wrong description for states


log1-c:
The current description for the "states" option in a dependency (template) tells you:

> The host/service states you want to get notifications for

According to the icinga2 docs this is wrong / quite the opposite:

> A list of state filters when this dependency should be OK. Defaults to [ OK, Warning ] for services and [ Up ] for hosts.

So if you configure your dependency, like the Director tells you, with "Critical" as state, from my understanding, you would get notified for all other service states but "Critical".

Also, on a side note:
The german translation
> ~~Die Host/Service **Stati,** für die Benachrichtigungen versandt werden sollen~~

should be

> Die Host/Service **Status**, für die Benachrichtigungen versandt werden sollen

[https://www.korrekturen.de/beliebte_fehler/stati.shtml](url) ;)

## Your Environment

* Director version (System - About): 1.4.3

log1-c:
I'll close this issue as I finally understood the thinking behind the wording for the states.
Typical "auf dem Schlauch stehen" from my side xD

PR #1481, opened 2018-05-07
Fix translation Stati -> Status

bargru:
Fix for https://github.com/Icinga/icingaweb2-module-director/issues/1446
see https://www.duden.de/rechtschreibung/Status


Thomas-Gelf:
Thanks!

PR #1484, opened 2018-05-09
php-diff: Fix undefined offset in Array Renderer

lazyfrosch:
See https://github.com/chrisboulton/php-diff/pull/51


Thomas-Gelf:
thnx!

Issue #274, opened 2016-06-10
[dev.icinga.com #11925] Do not allow Service States in Host Notifications


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/11925

**Created by chrkli193 on 2016-06-10 11:11:03 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-06-10 11:11:03 +00:00 (in Redmine)_

---

It is possible to add service states in a host notification. If you do that icinga2 got an error.

chas0rde:
ran into that previously as well
At best UI would prevent that

PR #1487, opened 2018-05-11
differentiate between host and service states in notification form

bargru:
With this pull request, only the relevant states for hosts OR services are displayed when a notification apply is created. Relates to #274

No Apply selected
grafik

Apply to Hosts
grafik

Apply to Services
grafik

Template:
grafik


bargru:
sing off went wrong. i will create a new pr...

Issue #1449, opened 2018-03-28
HostGroup assigned via HostTemplate is not considered for authentication role


zampat:
## Current Behavior / How to reproduce:
Defining in Icinga Authentication -> Roles a filter "director/filter/hostgroups" a host group.
Assign the hostgroup the hosts via template
The hosts having this hostgroup assigned for via host template are not visible in director

## Expected Behavior
The rights should be granted to hosts as if hostgroup is assigned directly to host.

Environment: Director 1.4.2

lazyfrosch:
Could you please share an example for objects and filters?

Is the assigned hostgroup shown beneath a host? (in the form should be "assigned hostgroups") - when browsing without restrictions.

zampat:
Hello Markus:

sure, here I show you a definition of a host group and a template having this host group assigned and the definition of a host, importing this hostgroup:

The Hostgroup definition used to grant monitoring view and edit rights in director:
zones.d/director-global/hostgroups.conf
object HostGroup "Windows Hosts Rome" {
}

The Host Template:
zones.d/director-global/host_templates.conf
template Host "generic-windows-rome" {
    check_command = "hostalive"
    max_check_attempts = "3"
    check_interval = 5m
    retry_interval = 1m
    check_timeout = 1m
    enable_notifications = true
    enable_active_checks = true
    enable_passive_checks = true
    enable_event_handler = true
    enable_perfdata = true
    volatile = false
    icon_image = "win.png"
    groups = [ "Windows Hosts Rome" ]
}

The Host Definition:
zones.d/master/hosts.conf

object Host "host_rome_dummy" {
    import "generic-windows-rome"
    address = "10.0.0.1"
}

The View in Director (logged in as Administrator): Yes, the Host Group appears to be assigend:
![image](https://user-images.githubusercontent.com/3853721/40602192-f666fb1c-6257-11e8-8824-d3682bcd9a8b.png)

Here I show how ROLE for administrator for Hosts of Hostgroup "Windows Hosts Rome" is defined:
![image](https://user-images.githubusercontent.com/3853721/40602243-2b481474-6258-11e8-9ae6-1cb8472d9a77.png)

When logging in as user "rome" I see the host in monitoring, but not in director:
![image](https://user-images.githubusercontent.com/3853721/40602304-60ec6f8a-6258-11e8-9247-aa05adf22eb3.png)


[...]

PR #1490, opened 2018-05-17
Membership cache not correctly updated in sync

lazyfrosch:
When object imports get changed by the sync.

The issue basically is, that the resolver is using the old imports, because `TemplateTree`
only uses the object presented when is hasn't been stored yet.

This always happens when the resolver is running deferred - after all objects have been stored.

Also I added a housekeeping CLI command to check and fix the cache:

```
$ icingacli director housekeeping memberships --apply
Checking host memberships
7825 objects - 1983 groups
Found 424 new and 56 outdated mappings
Update complete.
```

### DIFF
--- a/application/clicommands/HousekeepingCommand.php
+++ b/application/clicommands/HousekeepingCommand.php
@@ -5,6 +5,7 @@ namespace Icinga\Module\Director\Clicommands;
 use Icinga\Exception\MissingParameterException;
 use Icinga\Module\Director\Cli\Command;
 use Icinga\Module\Director\Db\Housekeeping;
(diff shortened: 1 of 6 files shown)


lazyfrosch:
I'd love to hear your opinion @Thomas-Gelf

Thomas-Gelf:
Related: #1449

lazyfrosch:
Rebased

lazyfrosch:
In light of #1618 this is slightly outdated.

* `getAncestorsFor()` will then work with new objects as well
* From this PR the housekeeping CLI should be extracted

TODO for me!

lazyfrosch:
Now incorporated into #1618

PR #1493, opened 2018-05-22
Sync: Correctly handle service templates and ignore apply

lazyfrosch:
* Using object_type apply is now explicitly prohibited
* Templates are correctly identified by name
* Sync will bail when existing objects don't have a named key

### DIFF
--- a/library/Director/Import/Sync.php
+++ b/library/Director/Import/Sync.php
@@ -395,14 +395,23 @@ class Sync
             $this->objects = array();
             $destinationKeyPattern = $this->rule->getDestinationKeyPattern();
 
+            $hasHostKey = strpos($destinationKeyPattern, '${host}') !== false;
+            $hasSetKey = strpos($destinationKeyPattern, '${service_set}') !== false;
+            $hasTypeName = strpos($destinationKeyPattern, '${object_type}') !== false;
+
             foreach (IcingaObject::loadAllByType(
                 $this->rule->object_type,
                 $this->db
             ) as $object) {
                 if ($object instanceof IcingaService) {
-                    if (strstr($destinationKeyPattern, '${host}') && $object->host_id === null) {
+                    if ($object->object_type === 'apply') {
+                        // we can not sync apply without a proper way of unique identification
                         continue;
-                    } elseif (strstr($destinationKeyPattern, '${service_set}') && $object->service_set_id === null) {
+                    } elseif ($hasHostKey && $object->host_id === null) {
+                        continue;
+                    } elseif ($hasSetKey && $object->service_set_id === null) {
+                        continue;
+                    } elseif ($hasTypeName && $object->object_type === 'object') {
                         continue;
                     }
                 }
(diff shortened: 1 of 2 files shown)


lazyfrosch:
Problem is that applied services are hard to identify, as they don't have a unique key, only the id.

`assign_filter` works well for Service Set or Groups.

With the current (before PR) feature set, you could create `apply Service` by Sync, but neither identify existing, nor update them. So every run would create new applies.

Regarding import source keys, they don't really affect the final key in Sync, as it is built from sync properties (even before my changes). The import source key here is mainly for ensured uniqueness.

lazyfrosch:
Btw. the current main problem here, is that you can't sync service templates at all... :cry:

lazyfrosch:
Part of this patch still needs to be merged - you currently can not reliable sync service templates.

I'll remove the numeric ID check and verify the patch asap.

lazyfrosch:
Updated this PR.

Tested syncing:
* Service templates
* Services for sets
* Services for hosts
* Unique apply Rules

lazyfrosch:
@Thomas-Gelf only real difference for the user should be the requirement for a combined key

Thoughts?

Thomas-Gelf:
Removing this from 1.6, IMO it's too hot. In case you want to have it in that version please let me know, we can have another look together.

lazyfrosch:
Well it's tested with large imports and configs.

Tell me what to test else...

Thomas-Gelf:
There was some forth and back here, we talked about this multiple times, not sure whether I'm up to date. My two main concerns where:

* how does this behave for Single Services from an ImportSource that does not explicitly define a combined key column pattern? I can remember that you told me that you addressed this, and the current pull request looks fine. Just to be sure that I didn't overlook anything: can you confirm this? (HIGHEST PRIO)
* does it still break Sync for Apply Rules where users attempted to deal with IDs in a hacky way? (LOWER PRIO)

To put it straight, my point of view on this is: Import/Sync is mainly for single objects. Hosts, Services, Users, Groups - also Datalist entries. These are the main use-cases and they are used intensively. This pull request touches services, and we should absolutely make sure that no existing setup breaks. To me, this has highest priority.

Synchronizing Templates or Apply Rules is technically possible, but has it's drawbacks. I always advised against doing so. Sure, I know that people are doing it anyways, and I personally helped some of them. Apply Rules are not unique, that's a big issue. Templates are mostly worthless without related fields, so to me Sync for them is flawed.

The Configuration Basket has been implemented to tackle this with v1.6. It intentionally does not support Service Apply Rules, but you can work around this by using applied Service Sets. In a future version we should tackle Apply Rules by replacing the Apply Rule ID with a UUID, combined with a UUID(5?)-Property-Modifier allowing you to UUIDize any source column you want.

In my understanding it's important for you to get Service Templates fixed, and that's perfectly fine for me. So please help me to get rid of my concerns, I'd love to merge this.

Thanks a lot for your patience,
Thomas

NB: I'd love to hear your thoughts related to the basket. Could it solve your requirement, what is missing?

lazyfrosch:
I added a patch that does the same for serviceset.

While I agree on doing a different approach, I'm using this on a user site for now.

I guess it would be better to enforce an `object_sub_type` on every sync, so you only can sync once each:
* Services for Host
* Services for Sets
* Service Templates
* Service Set Templates
* Service Set for Host

IMHO this PR should not break anything, but it might not be clear for users what to set.

-> So we would need integration tests and a proper documentation

PR #1522, opened 2018-06-13
Concatenates a the given strings before and/or after the value of the property

cstegm:
return $before.$value.$after;


Thomas-Gelf:
Is the `Combine` Property Modifier what you're looking for?

cstegm:
It is - the issue was that i thought that only fields can be combined. But I can also combine fields and normal text using the Combine Property Modifier.

PR #1557, opened 2018-07-04
Db/mysql: Hex binary blobs prior transmission and unhex them later

### LINKED ISSUE #1556: Binary blobs are not properly transmitted to MySQL
Given a `rowset_checksum` `�*���O��+d�C\"�I�` results in an invalid query string produced by Zend_Db_Statement:190 and ultimately in a segmentation fault caused by Zend_Db_Statement:196.

The invalid query is caused by the backslash in the checksum whereas the segmentation fault is a result of the invalid query. If the query is valid the segmentation fault does not occur.

So the actual issue is that the Director does not properly transmit binary blobs to MySQL.

## Expected Behavior
The director should properly escape binary blobs when querying the database.

## Current Behavior
In case of MySQL the director does not escape binary blobs at all.

## Possible Solution
Escape binary blobs or transmit them as part of prepared statements.

Just have a row with the above `rowset_checksum` and run a syncrule check.

## Your Environment
* Director version (System - About): master@35815e0d
* Icinga Web 2 version and modules (System - About): 2.5.3 (ca7e886b820f0145999fdf7377f5c4132301eada)

---
ref/NC/565488

nilmerg:
fixes #1556

### DIFF
--- a/library/Director/Db.php
+++ b/library/Director/Db.php
@@ -663,6 +663,8 @@ class Db extends DbConnection
     {
         if ($this->isPgsql()) {
             return new Zend_Db_Expr("'\\x" . bin2hex($binary) . "'");
+        } elseif ($this->isMysql()) {
+            return new Zend_Db_Expr('UNHEX(\'' . bin2hex($binary) . '\')');
         }
 
         return $binary;


lazyfrosch:
Closing this in favor of prepared statements

PR #1588, opened 2018-08-07
Add input suggestion for datalist vars

lazyfrosch:
Variables that have a datalist behind should be able to list suggestions.

This PR adds the appropriate suggestion lookup for all vars that are based on a datalist.

## Simple Text fields

bildschirmfoto von 2018-08-07 15-29-31

## Extensible sets

bildschirmfoto von 2018-08-07 15-29-48

ref/NC/565994

### DIFF
--- a/application/controllers/SuggestController.php
+++ b/application/controllers/SuggestController.php
@@ -19,9 +19,19 @@ class SuggestController extends ActionController
     {
         // TODO: Using some temporarily hardcoded methods, should use DataViews later on
         $context = $this->getRequest()->getPost('context');
+        $key = null;
+
+        if (strpos($context, '!') !== false) {
+            list($context, $key) = preg_split('~!~', $context, 2);
+        }
+
         $func = 'suggest' . ucfirst($context);
         if (method_exists($this, $func)) {
-            $all = $this->$func();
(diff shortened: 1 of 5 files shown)


lazyfrosch:
@Thomas-Gelf asked via E-mail (translated)

> Could you explain the change for the suggestion/autocomplete properties? I'm afraid we may loose some properties here. My browsers don't have a problem.
> And especially for extensible sets, are the properties still present there?
> What changes and what will be better?

In short the move to JS is to have a consistent handling for suggest/autocomplete=off. E.g. in the old code, autocomplete was not always disabled:

https://github.com/Icinga/icingaweb2-module-director/blob/14a3039f650f7978e4e3319ee5ce73851a4ce8d8/application/views/helpers/FormDataFilter.php#L231-L234

https://github.com/Icinga/icingaweb2-module-director/blob/14a3039f650f7978e4e3319ee5ce73851a4ce8d8/application/views/helpers/FormDataFilter.php#L411-L414

So Javascript will implicitly disable autocomplete for all fields we are suggesting on.

But you are right in terms of ExtensibleSet, there I removed one to many. Will fix this.

lazyfrosch:
Updated and rebased

PR #1597, opened 2018-08-10
RestoreObjectForm: Fix restore of multi-key and apply

lazyfrosch:
* Apply: Now uniquely identified via name (multiple matches result in error)
* Multi-Key: Loading by Multi-Key and validating object_type

Worst case an error would be thrown, but it should not overwrite the wrong object.

Tested with:
- [x] Service Set (template + object)
- [x] Service Apply
- [ ] Service (for host or template)
- [ ] Host (object + template)
- [ ] Notification
- [ ] Simple objects like user and user group

### DIFF
--- a/application/forms/RestoreObjectForm.php
+++ b/application/forms/RestoreObjectForm.php
@@ -2,7 +2,10 @@
 
 namespace Icinga\Module\Director\Forms;
 
+use Icinga\Exception\NotFoundError;
+use Icinga\Exception\NotImplementedError;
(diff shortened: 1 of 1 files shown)


lazyfrosch:
Please update the linked issues when merging

lazyfrosch:
What I noticed: Restoring a rename will just re-create the previous object (But I think that was before)

I updated the apply lookup, so it should work with other applied objects. (When they are object_type apply)

PR #1618, opened 2018-08-29
GroupMembershipResolver: Fix issues with apply and static assignments

### LINKED ISSUE #1574: Integrity constraint violation: icinga_hostgroup_host_resolved
After update from r2.8.4-1 to 2.9, all previous defined assign filters on HostGroups are no longer visible (but active) and an error message appears on saving assign filters on new and existing HostGroups.

## Expected Behavior
No error message should appear and all previous defined assign filters should be there.

## Current Behavior
Error message appears on save:
`SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '15-3' for key 'PRIMARY', query was: INSERT INTO icinga_hostgroup_host_resolved (hostgroup_id, host_id) VALUES (?, ?) (Pdo.php:225)`

## Possible Solution
NA

Create/edit a HostGroup and set/edit a assign filter

## Your Environment
* Director version (System - About): master
* Icinga Web 2 version and modules (System - About): 2.5.3
* Icinga 2 version (`icinga2 --version`): r2.9.0-1
* Operating System and version: Ubuntu 18.04 LTS
* Webserver, PHP versions: nginx/1.14.0, php7.2.7-1

lazyfrosch:
* Fixing static resolving in general
* Avoiding a problem where apply were matched when assign_filter is an
  empty string (empty FilterAnd)
* Adapt to ResolverTest

fixes #1574

### DIFF
--- a/library/Director/Objects/GroupMembershipResolver.php
+++ b/library/Director/Objects/GroupMembershipResolver.php
@@ -6,6 +6,7 @@ use Icinga\Application\Benchmark;
 use Icinga\Data\Filter\Filter;
 use Icinga\Module\Director\Db;
 use Icinga\Module\Director\Db\IcingaObjectFilterHelper;
+use Icinga\Module\Director\Repository\IcingaTemplateRepository;
 use InvalidArgumentException;
 use LogicException;
 use Zend_Db_Select as ZfSelect;
@@ -74,6 +75,11 @@ abstract class GroupMembershipResolver
     public function refreshDb($force = false)
     {
         if ($force || ! $this->isDeferred()) {
+            if ($this->isDeferred()) {
+                // ensure we are not working with cached data
+                IcingaTemplateRepository::clear();
+            }
+
             Benchmark::measure('Rechecking all objects');
             $this->recheckAllObjects($this->getAppliedGroups());
             Benchmark::measure('Recheck done, loading existing mappings');
--- a/library/Director/Objects/GroupMembershipResolver.php
+++ b/library/Director/Objects/GroupMembershipResolver.php
@@ -5,6 +5,7 @@ namespace Icinga\Module\Director\Objects;
 use Icinga\Application\Benchmark;
 use Icinga\Data\Filter\Filter;
 use Icinga\Module\Director\Db;
+use Icinga\Module\Director\Db\IcingaObjectFilterHelper;
 use InvalidArgumentException;
 use LogicException;
 use Zend_Db_Select as ZfSelect;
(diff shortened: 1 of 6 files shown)


lazyfrosch:
This addresses an issue where updates are mixed over the intended range of objects and groups. Mapping are updated where the respective groups are not in scope, and therefor can not be checked for "existing".

GroupMembershipResolver has 3 modes we use:
1. All groups for one object (changing a host)
2. All objects for one group (changing a host group)
3. All for all (e.g. after sync)

We should also add unit tests probably...

lazyfrosch:
Following up on this, my change here just excludes static hostgroups from the resolver.

Working on a better solution, including tests

lazyfrosch:
This has become a quite large diff, but still only bugfixing issues with the resolver.

The Test included should test the basic behavior when changing objects.

@Thomas-Gelf please review!

lazyfrosch:
There seems to be a problem with PHP < 5.6 - need to verify this

lazyfrosch:
Included housekeeping and syncmode from #1490

lazyfrosch:
Removed a mis-commited file from testing

lazyfrosch:
Current testing this in a larger environment.

Still having some problems with group assign -> host when going by vars, and the var is inherited from a template.

Hostgroup has been updated by sync, a housekeeping call didn't fix it, but manual change of the group

lazyfrosch:
Tracked this down to a rogue custom var, mapping seem better now, but still testing.

Thomas-Gelf:
@lazyfrosch: I tried a simplistic approach for just a single use-case, #1574 - could you please have a quick look at 131a921 (temp. branch: fix/hostgroup-membership-resolver)?

Thomas-Gelf:
@lazyfrosch: would splitting this into "single group updates like in #1574", "massive changes in Sync" and "housekeeping utility for group membership cache" cause a problem for your working environment?

lazyfrosch:
Removed all commits that are for testing and housekeeping.

We will add this to 1.6 afterwards

library/Director/Objects/GroupMembershipResolver.php
```diff
@@ -392,8 +431,14 @@ protected function addMembershipWhere(ZfSelect $query, $column, & $objects)
         }
 
         $ids = array();
-        foreach ($objects as $object) {
-            $ids[] = (int) $object->get('id');
+        foreach ($objects as $k => $object) {
+            if (is_string($object)) {
+                $ids[] = (int) $k;
+            } elseif (is_numeric($object)) {
```
Thomas-Gelf: Please change  this to is_int, remove (int) in the line below - then merge to master

lazyfrosch: done

lazyfrosch:
Cherry-Picked to `support/1.5`

Thomas-Gelf:
@lazyfrosch: sh**... discovered a problem: this now writes also matching templates to the DB... was that intentional?

PR #1620, opened 2018-08-31
Add function for cache the template tree status.

ValeDaRold:
The tree you can find in host/service templates will be reloaded according to the autorefresh.
The user will loose the status of the tree while he is working on it. For example this is how the user sets the tree, while searching something through the tree entries:
image
After 10 sec. the autorefresh will make the page reloads like:
image
In order to avoid this annoying behavior, I created a javascript function that saves the status of the tree in cache and loads the user configuration each time the page is refreshed.


Thomas-Gelf:
Sorry for getting back to you so late on this one. Reason for skipping this for now: the url-based localstorage solution feels a little bit hacky, especially given that we already have identifiable window identifiers and session namespaces per browser tab.

PR #1623, opened 2018-09-05
IcingaServiceSet: Render sub-sets in config preview

lazyfrosch:
This allows the user to preview all sub-service sets.

Normally the preview only shows
1. the applied services
2. when only directly added service-sets it is empty

bildschirmfoto von 2018-09-05 12-30-01
bildschirmfoto von 2018-09-05 12-29-50

### DIFF
--- a/library/Director/Objects/IcingaServiceSet.php
+++ b/library/Director/Objects/IcingaServiceSet.php
@@ -2,6 +2,7 @@
 
 namespace Icinga\Module\Director\Objects;
 
+use Exception;
 use Icinga\Data\Filter\Filter;
 use Icinga\Module\Director\Exception\DuplicateKeyException;
 use Icinga\Module\Director\IcingaConfig\IcingaConfig;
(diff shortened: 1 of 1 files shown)


library/Director/Objects/IcingaServiceSet.php
```diff
@@ -327,4 +327,29 @@ protected function beforeStore()
             );
         }
     }
+
+    public function onSingleIcingaConfig(IcingaConfig $config)
```
Thomas-Gelf: I do not like the name of this method. As you're calling parent:: anyways, couldn't you just override `toSingleIcingaConfig()` to get the same result? It would duplicate the try/catch part, but a) who cares or b) move that part to a `renderFailsafe($callback)` method.

library/Director/Objects/IcingaServiceSet.php
```diff
@@ -327,4 +327,29 @@ protected function beforeStore()
...
+    {
+        parent::onSingleIcingaConfig($config);
+
+        if ($this->id !== null) {
```
Thomas-Gelf: Please be careful, `$this->id` is a very special beast, it will probably end up calling `getId()`. It might work here, but I'd prefer to see the more explicit `get('id')` everywhere. This is part of some black magic in `DbObject` I absolutely do not like, I wrote id long long time ago - it scares me ;-)

lazyfrosch:
I guess the hook is unnecessary, updated to your suggestions.

Thomas-Gelf:
Thanks!

PR #1624, opened 2018-09-05
Improving legacy rendering

lazyfrosch:


Thomas-Gelf:
Is this ready to be merged? Looks good to me, and apart from some base object features that I'd love to externalize later on this looks as it wouldn't break anything 2.x related. In case it isn't please consider merging the fix for flattenVars to master in the meantime.

Thanks,
Thomas

lazyfrosch:
This should be fine and not break any behavior.

Issue #1621, opened 2018-09-03
Zone configuration overriden when using apply rules


matthenning:
I encountered an issue when trying to deploy a service with a defined zone using an apply rule.

## Expected Behavior
If I define a zone for a service and apply it using an apply rule in Director the service is deployed accordingly and is executed in the defined zone even when applied to hosts in other zones.

## Current Behavior
The zone is overriden and the service stays `pending` forever. Config checks succeed.
This seems to occur if either the apply rule or the template is configured in director. If thy are both defined in /etc/icinga2/zones.d/master it works.

Director template preview:
```
zones.d/master/service_templates.conf

template Service "REMOTE ES Query Automic Broken Pipe" {
    check_command = "check_elasticsearch_query"
    max_check_attempts = "3"
    check_interval = 30s
    retry_interval = 30s
    check_timeout = 30s
    enable_notifications = true
    enable_active_checks = true
    enable_passive_checks = true
    enable_event_handler = true
    enable_perfdata = true
    zone = "master"
    icon_image = "automic.png"
    command_endpoint = null
}
```

### When using the apply rule it's broken:

Director Apply Rule preview:
```
apply Service "App - Automic Agent Broken Pipes" {
    import "REMOTE ES Query Automic Broken Pipe"

    assign where host.vars.app_automic && host.vars.app_automic_db

    import DirectorOverrideTemplate
}
```

Icinga 2 object:
```
# icinga2 object list --type Service --name "App - Automic Agent Broken Pipes"
Object 'REDACTED!App - Automic Agent Broken Pipes' of type 'Service':
  % declared in '/var/lib/icinga2/api/packages/director/4e72a767-6b75-4fff-b2d6-1b55f2066b4b/zones.d/master/service_apply.conf', lines 1:0-1:47
  /** truncated **/
  * zone = "satellite"
    % = modified in '/var/lib/icinga2/api/packages/director/4e72a767-6b75-4fff-b2d6-1b55f2066b4b/zones.d/master/service_apply.conf', lines 1:0-1:47
```

The line where the zone is modified just contains the apply rule I pasted above.

### When assigning directly to a host it works:

Director Service preview:
```
object Service "App - Automic Agent Broken Pipes" {
    host_name = "REDACTED"
    import "REMOTE ES Query Automic Broken Pipe"

}
```

Icinga 2 object:
```
# icinga2 object list --type Service --name "App - Automic Agent Broken Pipes"
[...]

PR #1634, opened 2018-09-05
WIP: IcingaService: enforce zone for apply rules

Thomas-Gelf:
refs #1621

### DIFF
--- a/library/Director/Objects/IcingaService.php
+++ b/library/Director/Objects/IcingaService.php
@@ -480,14 +480,24 @@ class IcingaService extends IcingaObject implements ExportInterface
 
     /**
      * @return string
+     * @throws \Icinga\Exception\NotFoundError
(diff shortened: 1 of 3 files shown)


Thomas-Gelf:
@lazyfrosch: this is what we have been talking about. Final patch will differ, I need to ask each time whether a Zone is global, one more information that requires caching. Also, I guess I'll move it to the base IcingaObject as this very same logic fits every Apply Rule.

For now I'm only interested in your opinion on the rendered config. Does it work for you? Can you imagine any scenario where this would not be what one expects?

lazyfrosch:
Hmm, I need some time to think about the possible outcomes of this change.

I'm not so sure about the generic user understanding how this would work, and when it does not...

Thomas-Gelf:
Well, they shouldn't think at all, that's the point :laughing: They'll have no influence on this, it should just work as expected.

PR #1654, opened 2018-09-27
Render Icinga config helpers with safe globals access

dnsmichi:
globals[key] works, although it wasn't documented how it
behaves in the case when this field doesn't exist.

Previously this has been a dictionary which silently returns
null if the key doesn't exist. With v2.10, globals turned
into a namespace which throws an error if not existing (which is
the correct behaviour).

In order to avoid configuration errors with 2.10, this is changed
into `contains()` which has been sitting in dictionaries since 2014.
So this patch ensures compatibility with older versions as well.

### DIFF
--- a/library/Director/IcingaConfig/IcingaConfig.php
+++ b/library/Director/IcingaConfig/IcingaConfig.php
@@ -534,7 +534,7 @@ globals.directorWarnOnceForThresholds = function() {
         return sprintf(
             '
 const DirectorOverrideTemplate = "%s"
-if (! globals[DirectorOverrideTemplate]) {
+if (! globals.contains(DirectorOverrideTemplate)) {
   const DirectorOverrideVars = "%s"
 
   globals.directorWarnedOnceForServiceWithoutHost = false;


Thomas-Gelf:
Looks good to me, thanks for the pull request and the related details. 2014 means that Agents running 2.6 are safe I guess. @lazyfrosch: if you have the chance to give 2.6 a quick test this would be cool, otherwise I'd blindly trust @dnsmichi and merge to master.

dnsmichi:
Anything since 2.4 should work here, `contains()` as dictionary method was available when `globals` was introduced the first time (December 2014, or so).

Thomas-Gelf:
Thanks again!

dnsmichi:
Since this problem comes out more often, I've discussed it again with @lippserd ... the patch for the Director is safe, still we need to take care about existing deployments where this would stop with 2.10. The patch for the Director is fine and safe, but we don't need a release for it now. I'll like a PR for 2.10 which solves the access problem, and allows for smooth upgrading.

PR #1660, opened 2018-10-02
Property Modifier Trim

cstegm:
Icinga2 and the Director allows to configure unwanted hostnames like " myhostname" wich leads often to issues connecting agents etc, so i need an easy to use propertymodifier for this.


Thomas-Gelf:
No feedback, I'll push a variant of this

PR #1678, opened 2018-10-16
Add URL to service template creation note

dnsmichi:
This follows the same principle as with hosts where the URL
already is put into. I've found it more convenient to just
click on it and not having to navigate into the services tree :)


dnsmichi:
Thanks for the explanation, I am always learning :-)

PR #1698, opened 2018-11-12
Added information about not importing commands issue.

xtavras:
I had problem that commands were not imported after kickstart in icingaweb, even after deleting endpoint and adding it again. The solution was to run kickstart manually via CLI. I thought this information will be helpful the others.

### DIFF
--- a/doc/80-FAQ.md
+++ b/doc/80-FAQ.md
@@ -59,6 +59,17 @@ it could happen that the default conservative max package size of your MySQL
 server bites you. Raise `max packet size` to a reasonable value, this willi
 usually fix this issue.
 
+Commands are not imported with Kickstart
+-------------------------------
+
+If after running kickstart you can't see imported check commands, try to run kickstart process manually from CLI.
+
+```sh
+icingacli director kickstart run
+```
+See "03-Automation.md" for more information.
+
+
 Import succeeded but nothing happened
 -------------------------------------
 


Thomas-Gelf:
@xtavras: what Icinga 2 version are you running? There is a known problem with Icinga 2 >= 2.8.2 and < 2.10.2, Director v1.6.0 should ship a workaround for this.

xtavras:
Hi @Thomas-Gelf 

icinga2 = version: r2.10.1-1
director = 1.5.2

Thomas-Gelf:
@xtavras: then you might be affected by a Bug in Icinga, introduced by some security measures in 2.8.2. This will be fixed in 2.10.2 - and Director 1.6 will ship a workaround. I'll not merge your pull request, as I fear that it would only increase confusion.

xtavras:
@Thomas-Gelf no objections, but I still would like to see mention of "kickstart run" CLI command somewhere in documentation for possible troubleshooting. I've wasted almost 3 hours trying to figure out why checks were not imported and found out about the command on some random forum post for different problem. I've only found it in "Automation.md" file by greping repo before this PR. I can rewrite it to be more generic and make new PR. 
What do you think?

Thomas-Gelf:
@xtavras: kickstart on CLI works differently than Kickstart in the UI. Usually you need a dedicated configuration file - something that normally doesn't exist. It's there solely for fully automated setups, and that's why it is documented in that section. This should NOT be proposed as a workaround because it will usually NOT work out of the box.

I'm aware of the underlying issue. Icinga 2 failing with more consecutive requests in the same HTTP session. And yes, it is hard to troubleshoot. This has been fixed in Icinga, it will be covered by 2.10.2. And Director 1.6.0 will ship a workaround, it will disconnect from the API after each request for every Core version >= 2.8.2 and < 2.10.2.

This does not happen all the times, depends on various factors and it is not granted that it will always work on CLI. Your proposed workaround might also fail.

Thomas-Gelf:
@xtavras: Icinga v2.10.2 has been released, a workaround for older versions is in the current Director master.

PR #1721, opened 2018-11-27
Allow ArrayByElementPosition modifier to filter by key name

joni1993:
Allows accessing array elements of associative arrays by key name

This allows for example to parse a JSON array like
```
{
  Monitoring: True,
  Monitoring Endpoint: "<TEXT>",
  Location: "<TEXT>",
}
```
to be decoded into an object and then to make a property out of one of the JSON array elements (e.g. for later filtering in Sync Rule)


jwie75:
I wanted to make exactly the same change proposal for this modifier,
luckily in open pull request found this here.

Have tested these changes successfully.

- icingaweb2-module-director: 1.6.0
- icingaweb2-module-aws: 0.6.0

This is exactly what we need to import from AWS Cloud for the EC2 tags.

Can I help somehow to make this happen faster?

getjack:
@Thomas-Gelf @lazyfrosch Would you maybe approve this PR or tell us what is missing?

Thomas-Gelf:
Looks good to me, thank you!

PR #1730, opened 2018-12-10
DbObject: Don't confuse object_names and ids

### LINKED ISSUE #1729: Creating objects with some auto increment id = object_name still has issues
Basic info see issue #992, issues are basically the same, though with different causes. In effect, we still had a patch lying around that fixed further issues with auto increment id = object_name.

This issue is for opening the respective pull request only.

rattacresh:
fixes #1729


Thomas-Gelf:
@rattacresh: thanks for this pull-request, and sorry for the late response. Main reason for not merging this one was that it changes an old fragile monster class, and it's not that easy to immediately see whether it would have side-effects.

In my believes the current master no longer shows the problem you were trying to fix. Could you please give it a try and let me know?

Thanks a lot,
Thomas

Thomas-Gelf:
ping @rattacresh

PR #1741, opened 2018-12-14
Import: Improve exception info during storing rows

lazyfrosch:
ref/NC/589845

### DIFF
--- a/library/Director/Import/Import.php
+++ b/library/Director/Import/Import.php
@@ -301,12 +301,20 @@ class Import
             }
 
             foreach ($newRows as $row) {
-                $db->insert('imported_row', $rows[$row]);
-                foreach ($this->rowProperties[$row] as $property) {
-                    $db->insert('imported_row_property', array(
-                        'row_checksum'      => $this->quoteBinary($row),
-                        'property_checksum' => $property
-                    ));
+                try {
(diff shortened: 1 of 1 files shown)


library/Director/Import/Import.php
```diff
@@ -301,12 +301,20 @@ protected function storeRowset()
...
+                    foreach ($this->rowProperties[$row] as $property) {
+                        $db->insert('imported_row_property', array(
+                            'row_checksum'      => $this->quoteBinary($row),
+                            'property_checksum' => $property
+                        ));
+                    }
+                } catch (Exception $e) {
+                    throw new IcingaException(
+                        "Error while storing a row for '%s' into database: %s",
+                        $rows[$row]['object_name'],
```
Thomas-Gelf: Do we always have an object_name here? If not, please handle that case too. The rest looks good!

lazyfrosch: Yes we do:

https://github.com/Icinga/icingaweb2-module-director/blob/58c06890052a07db57ab8d3d26c481e08b3de4e8/library/Director/Import/Import.php#L262-L265

Thomas-Gelf: Perfect, thanks!

PR #1746, opened 2018-12-21
Add PropertyModifierParseURL.

dawnbreak:
Provide a PropertyModifier for the PHP function parse_url().

I have a lot of websites that I need to monitor which come from different sources. This PropertyModifier is quite handy for me. Maybe also for others?


Thomas-Gelf:
Well done, thank you!

PR #1752, opened 2019-01-04
Add description field to command argument form

hrak:
This small PR adds a form field to the command argument form that allows the user to add a description to a command argument.

Everything on the API side already seems to be in place and i was able to test it successfully (adding a description to an argument makes it show up in the command preview).


Thomas-Gelf:
Thank you!

PR #1759, opened 2019-01-14
RestApiClient#curlRequest(): remove timeout

Al2Klimov:


Al2Klimov:
Unfortunately Icinga 2 sometimes fails to complete the TCP handshake soon enough.

Thomas-Gelf:
@Al2Klimov: if I recall it correctly you get a red error message and will be allowed to deploy again when Icinga fails to respond. 3 seconds are an eternity in a modern datacenter, especially as we're talking about the connection time and not about the time required for a full deployment.

Setting this to the default would probably result in a timeout of  300 seconds. After this time many users already run into FPM or other timeouts - and get no useful error at all. And even if he would still get our error response: we cannot expect a web user to wait that long unless we tell him that a connection timed out.

@lippserd: if you see a requirement for this please let's discuss this elsewhere.

Al2Klimov:
To be honest I just wanted to buy some time for https://github.com/Icinga/icinga2/issues/6886 or similar. But if both of you disagree...

PR #1766, opened 2019-01-23
Unify dictionary key rendering

Al2Klimov:
... to protect the user against Icinga 2's breaking changes.


Al2Klimov:
Didn't know that Icinga 2 added any reserved keywords and that many of them...

Thomas-Gelf:
I consider the current rendering easier to read and closer to all examples shown in our documentation. If we can assume that Icinga 2 mentions breaking changes in it's release notes it shouldn't be a big thing to catch up. Especially as breaking config format changes should be an exemption and not the rule.

Thanks for your efforts, I'll not merge this pull request.

PR #1770, opened 2019-02-01
feature/notification_interval

n0braist:
it would be great if there would be the possibility using a custom host or service variable instead of a fixed value in the notifation definitions for the field notification_interval.
The background is, that so it would be possible, that one 'global' notification rule can be used with different renotifications instead of recreating and not assigned rules for a few individual notifications.

### DIFF
--- a/application/controllers/ConfigController.php
+++ b/application/controllers/ConfigController.php
@@ -313,7 +313,12 @@ class ConfigController extends ActionController
     public function storeAction()
     {
         $this->assertPermission('director/deploy');
-        $config = IcingaConfig::generate($this->db());
+        try {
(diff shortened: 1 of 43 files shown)


widhalmt:
ref/NC/603133

lazyfrosch:
looks like this PR has the wrong branch set to compare.

We won't merge `support/1.6` into `master`.

Is this meant as issue, or do you have a branch with changes somewhere I can't see?

n0braist:
Hi,

vermutlich war ich einfach nur zu dusselig es „normalen“ feature-request einzutüten….

Grüße Stefan

lazyfrosch:
Closing this PR then :wink:

PR #1773, opened 2019-02-06
IcingaService: Fix getOnDeleteUrl() for service sets

lazyfrosch:
So that it redirects back to services of this serviceset.


lazyfrosch:
Backported for 1.6.1

PR #1786, opened 2019-02-13
IcingaServiceSet: Delete host sets when deleting a set

lazyfrosch:
This allows service sets to be deleted, including their children sets that
are assigned to hosts.

The UI forbids deleting those sets, but we can use it like this for CLI
and purging during sync.


Thomas-Gelf:
@lazyfrosch: good catch! Did you already check how this behaves in the Activity Log, or shall I do so? I'd then consider merging this into v1.6.1, scheduled for tomorrow - any objections?

lazyfrosch:
Apart from rendering in activity log everything works fine.

PR #1796, opened 2019-03-05
Allow IcingaService to be cloned to a different host

lazyfrosch:
Similar to the implementation for Service Set.

bildschirmfoto von 2019-03-05 12-03-07


lazyfrosch:
I guess select is here a bit slow for large lists of host.

lazyfrosch:
This has been refactored to suggestion input:
Bildschirmfoto von 2019-03-19 15-00-39

Thomas-Gelf:
Well done, thank you!

PR #1806, opened 2019-03-21
ConfigHelper: Ensure empty vars are not written as void to config

lazyfrosch:
In renderStringWithVariables()

This can happen when variables are stored with an empty string, for an apply rule.

Normally Director won't save empty string variables, it will delete the var instead. But when you set all vars as array via sync.

If a user would have synced this to vars, it was rendered without any quotes.

```json
{
  "a": "test",
  "b": ""
}
```

Until now:

```
    vars.b =
```

Fixed:

```
    vars.b = ""
```

### DIFF
--- a/library/Director/IcingaConfig/IcingaConfigHelper.php
+++ b/library/Director/IcingaConfig/IcingaConfigHelper.php
@@ -406,6 +406,10 @@ class IcingaConfigHelper
             $parts[] = static::renderString(substr($string, $offset, $i - $offset));
         }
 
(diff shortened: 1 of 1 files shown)


Thomas-Gelf:
Hi @lazyfrosch,

until now, empty strings are not a thing in the Director - there is no way to set them in the GUI. If you add a field for your variable, it might get removed when storing the object next time. The fix itself is correct, but gives the wrong impression that empty strings are supported. Should we have a quick discussion about this?

Cheers,
Thomas

PR #1809, opened 2019-03-22
IcingaCommand: Allow command to be rendered as string

lazyfrosch:
Icinga 2 allows this from the beginning.

This is needed for migration purposes and old commands.

- [x] Add a security warning to the description of the object attribute

ref/NC/573707

### DIFF
--- a/application/forms/IcingaCommandForm.php
+++ b/application/forms/IcingaCommandForm.php
@@ -76,8 +76,9 @@ class IcingaCommandForm extends DirectorObjectForm
 
         $descIsString = [
             $this->translate('Render the command as a plain string instead of an array.'),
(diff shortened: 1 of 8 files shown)


lazyfrosch:
Here is a brief example of rendered config:

```
object CheckCommand "check_by_ssh" {
    import "plugin-check-command"
    import "icinga-legacy-check"

    command = PluginDir + "/check_by_ssh -l icinga -H $host.address$ -E 16 -t 60 -o \"StrictHostkeychecking=no\" -C \"$vars.ARG1$\""
}
```

lazyfrosch:
We will only add a warning to the object attributes description for now...

PR #1817, opened 2019-03-28
Specify command endpoint on service template

pbngl:
Gives you the option to specify command endpoint on a service template. Uses drop down list of all defined endpoints in director table.

Storing configuration sets the command_endpoint_id in director.icinga_service

### DIFF
--- a/doc/02-Installation.md
+++ b/doc/02-Installation.md
@@ -84,7 +84,7 @@ You might want to use a script as follows for this task:
     ICINGAWEB_MODULEPATH="/usr/share/icingaweb2/modules"
     REPO_URL="https://github.com/icinga/icingaweb2-module-director"
     TARGET_DIR="${ICINGAWEB_MODULEPATH}/director"
-    MODULE_VERSION="1.6.0"
+    MODULE_VERSION="1.6.2"
     URL="${REPO_URL}/archive/v${MODULE_VERSION}.tar.gz"
     install -d -m 0755 "${TARGET_DIR}"
     wget -q -O - "$URL" | tar xfz - -C "${TARGET_DIR}" --strip-components 1
@@ -101,7 +101,7 @@ It will be immediately ready for use:
     ICINGAWEB_MODULEPATH="/usr/share/icingaweb2/modules"
     REPO_URL="https://github.com/icinga/icingaweb2-module-director"
     TARGET_DIR="${ICINGAWEB_MODULEPATH}/director"
-    MODULE_VERSION="1.6.0"
+    MODULE_VERSION="1.6.2"
     git clone "${REPO_URL}" "${TARGET_DIR}" --branch v${MODULE_VERSION}
 
 You can now directly use our current GIT master or check out a specific version.


magenbrot:
This is exactly what I need. I would love to see this PR merged :+1:

Thomas-Gelf:
Not merging right now, this is dangerous. I see the need for this, but this is an easy way to shoot yourself in your knee. I fear that as soon as we offer this possibility, quite some people will fill this field on all of their default templates with "default values". We should present this together with a big warning or similar.

raidsec:
When could we use this function?

skupjoe:
Aren't there DB schema changes that need to be made for this to fully work? Currently, there is a boolean stored for `use_agent` column in the `icinga_service` table. Wouldn't there need to be a `command_endpoint_id` column added?

khmuth:
I fully agree with Thomas-Gelfs comment. #issuecomment-533354038
Therefore I prefer the possibility to set the `command_endpoint` only in the service definition and not in the service template to prevent the "shot in the knee".

skupjoe:
I think the benefits for users requiring this feature which is supported in Icinga2 but not Director currently outweigh the possible risks to new users that can be mitigated in the same way that zone selection is provided for other configurations, with proper warning messages & verbiage. Let's cater to our advanced users and unlock the full functionality of Director.

skupjoe:
Use case: I have isolated environments for my customer vms. For these hosts, I have vmware checks for some services that need to be triggered by one endpoint, while I have OS-level checks being performed by SNMP that need to be triggered by a separate endpoint in a different zone.

picnicsecurity:
For those that come after me, this still works as of 12/10/2020. Details below

### Use case
I need remote checks to be done by the Satellites and not the Master. The client endpoints I am monitoring just need simple remote ping/http/ssh/winrm checks done. Host templates allows me to specify the command endpoint which allows me to do ping checks through hostalive just fine. However I cannot check other services like ssh.

### How I added the file to my master endpoint
The first thing I did was stop the icinga2 service and the apache2 service. I then dropped this file into `/usr/share/icingaweb2/modules/director/application/forms/IcingaServiceForm.php` after backing up the one that was already there. I then restarted icinga and confirmed it started fine. Then I started apache2 and made sure it came up fine. Afterwards I went into Director and confirmed that I could select a satellite as the command endpoint for a service template. Everything has worked fine so far.

### Relevant Issue 
https://github.com/Icinga/icingaweb2-module-director/issues/2250
I have opened an issue about this because I am not sure if there is a "supported" way to achieve what I want with the current stable version of Director. I will update this issue with any of my findings.

### Icinga Env
image


Thank you @pbngl !!!

picnicsecurity:
I was able to finally figure this out thank to Roland Sommer on the Icinga Community forms.

In order to make this work the way I want to I need to not set the Cluster Zone, Icinga2 Agent, or Command Endpoint values on the service and host templates. They should be left as `-Please Choose-`. On the host itself however, that is where I set the Cluster Zone to the site that it is in, and if it is a remote check like the ones that I am doing, I set the Icinga2 Agent field to No.  This means I won't have to have a service check for each site and I wont need the above mod.

RafaelSche:
Any news on when or if this will be merged? That would be really nice.

chucker:
Does this mean that, with Icinga Director, I _cannot_ have a service whose command endpoint is the agent?

If so, what is the workaround?

slalomsk8er:
> Does this mean that, with Icinga Director, I _cannot_ have a service whose command endpoint is the agent?
> 
> If so, what is the workaround?

It means, that currently only the master or the agent can be the endpoint.

PR #1878, opened 2019-05-22
Ensure deployment logs are always collected

lazyfrosch:
Especially before running health checks.

Main problem with this was that for the user in UI, logs where only collected when reloading the deployment list.

If the user doesn't wait long enough, by switching to a different controller, the deployment stays stale.

Same can happen when the ConfigJob deploys, and the deployment does not finish within 3 seconds.

@Thomas-Gelf opinion?

### DIFF
--- a/library/Director/Health.php
+++ b/library/Director/Health.php
@@ -2,6 +2,7 @@
 
 namespace Icinga\Module\Director;
 
+use Exception;
 use Icinga\Application\Config;
 use Icinga\Date\DateFormatter;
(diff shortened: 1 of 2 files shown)


Thomas-Gelf:
@lazyfrosch: IMHO running a health check should trigger no action at all, without exemptions. The background daemon is going to be mandatory anyways with v1.7.0 - this can become one of it's first tasks. I'd love to finally remove this completely from the front-end.

Thomas-Gelf:
Created #1905, closing this one.

PR #1887, opened 2019-05-31
doc: add a note about director-global being reserved for Director

widhalmt:
We had some users using `director-global` for syncing their own configuration files. While this is technically possible it's not "best practice" and can produce some side effects. Therefore the warning.


Thomas-Gelf:
Thanks!

Issue #1142, opened 2017-08-30
dependencies: unable to resolve related property


vita2:
i'm trying to create the dependencies examples from the icinga2 documentation ([Dependencies for Network Reachability](https://www.icinga.com/docs/icinga2/latest/doc/03-monitoring-basics/#dependencies-for-network-reachability)) but i can't choose the necessary service as parent service.

## Current Behavior

currently my host `pfsense.home.lan` has two services:
* http-cert
* ping4

both services will be applied with help of assignment rules:
```
apply Service "http-cert" {
    import "http"

    assign where host.name == "pfsense.home.lan"
    vars.http_certificate = 31
    vars.http_ssl = true

    import DirectorOverrideTemplate
}

apply Service "ping4" {
    import "ping4"

    assign where host.address

    import DirectorOverrideTemplate
}
```

at the dependencies form i can only choose the service `http-cert` but i need the `ping4`-service as parent service. i think the field can't find services which was assigned by the new "is true/set" operator.

![add-dependency-01](https://user-images.githubusercontent.com/25298055/29865821-d40e202e-8d76-11e7-9e91-5046dc99bbf5.png)

when trying to input the `ping4` service manually i will get the error message:
```
Unable to resolve related property: "parent_service_id" (IcingaDependency.php:414)
```
![add-dependency-02](https://user-images.githubusercontent.com/25298055/29865822-d41031f2-8d76-11e7-906c-59385f28dae9.png)

## Steps to Reproduce (for bugs)
1. create a ping4 service template
2. apply the template to all hosts with "is set" operator (`assign where host.address`)
3. try to create a dependency and choose the new ping4 service as parent service

## Your Environment
* Director version (System - About): 604cd245cdc44a234b6cc0dc6ad344229d79e8b9 feature/dependencies-1096
* Icinga Web 2 version and modules (System - About): 2.4.1
* Icinga 2 version (`icinga2 --version`): r2.7.0-1
* Operating System and version: Debian Jessie 8.9 (3.16.0-4-amd64)
* Webserver, PHP versions: nginx 1.6.2-5 + php5-fpm 5.6.30

Thomas-Gelf:
@vita2: I merged dependencies to master. They are still considered experimental, but should work fine. The issue you've raised should no longer exist, as "single dependency objects" are currently not allowed, just apply rules. In case you're still experiencing related issues please let me know.

log1-c:
Hi @Thomas-Gelf ,

[...]

PR #1906, opened 2019-07-06
fix for issue #1142

terra-nova:
Establishing service-to-service dependencies always fails with an error message stating that the `parent_service_id` propery could not be resolved. I believe this is due to a misnamed dictionary key (probably some kind of copy&paste error?), although I admit I do not fully comprehend the inner workings of Icinga Director.


Thomas-Gelf:
@terra-nova: thanks for this, but for me this doesn't really fix the problem. It seems that the original author of this piece of code had something different in mind, I'm currently trying to understand what that might have been.

The condition you're changing is correct, that piece of code should only run if no parent host(!) had been chosen. So this is probably for dependencies against a specific Service on the very same host.

PR #1934, opened 2019-09-02
Fix bad looking form errors for Web 2 >= 2.7.0

lippserd:
Starting with Web 2.7.0 the reworked form styles are only available for
for forms with the CSS class icinga-form. Since the Director uses its
own form styles, the styles for form errors no longer apply and thus
look ugly. This commit (re)introduces those rules for the Director.

Before:

localhost_8080_icingaweb2_config_modules (1)

After:

localhost_8080_icingaweb2_director_settings

### DIFF
--- a/public/css/module.less
+++ b/public/css/module.less
@@ -330,7 +330,14 @@ input[type=text], input[type=button], select, select option, textarea {
 }
 
 form ul.form-errors {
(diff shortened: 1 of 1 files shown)


Thomas-Gelf:
Thanks!

Issue #1535, opened 2018-06-19
Director want to reencode DB to UTF8 but its already done


ep4sh:
## Expected Behavior
I see in Icinga Director -> Overview (near DB resourse):

> Please change the encoding for the director database to utf8

Also, when i try to add some template or othe item:

> Storing director_activity_log[] failed: SQLSTATE[22021]: Character not in repertoire: 7 ОШИБКА: неверная последовательность байт для кодировки "UTF8": 0xd0 0x32, query was: INSERT INTO director_activity_log (object_name, action_name, object_type, old_properties, new_properties, author, change_time, checksum, parent_checksum) VALUES (?, ?, ?, ?, ?, ?, ?, '\xd02c30f8d52aceb5bd6d622378855d4c2148cf04', '\x') {array ( 'id' => NULL, 'object_name' => '21', 'action_name' => 'create', 'object_type' => 'icinga_host', 'old_properties' => NULL, 'new_properties' => '{"object_name":"21","object_type":"template"}', 'author' => 'icinga2', 'change_time' => '2018-06-19 14:22:50', 'checksum' => '�,0��*ε�mb#x�]L!H�', 'parent_checksum' => NULL, )} (DbObject.php:824)

Also i created my db using doc pages:
`psql -q -c "CREATE DATABASE director WITH ENCODING 'UTF8';"`

## My env
```
* Director version (System - About):
director | master
```

Icinga Web 2 Version    2.5.3
* Icinga 2 version (`icinga2 --version`):

```
icinga2 - The Icinga 2 network monitoring daemon (version: r2.8.4-1)
Copyright (c) 2012-2017 Icinga Development Team (https://www.icinga.com/)
License GPLv2+: GNU GPL version 2 or later <http://gnu.org/licenses/gpl2.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Application information:
  Installation root: /usr
  Sysconf directory: /etc
  Run directory: /run
  Local state directory: /var
  Package data directory: /usr/share/icinga2
  State path: /var/lib/icinga2/icinga2.state
  Modified attributes path: /var/lib/icinga2/modified-attributes.conf
  Objects path: /var/cache/icinga2/icinga2.debug
  Vars path: /var/cache/icinga2/icinga2.vars
  PID path: /run/icinga2/icinga2.pid

```
> System information:
>   Platform: Debian GNU/Linux
>   Platform version: 9 (stretch)
[...]

PR #1935, opened 2019-09-03
KickstartForm: add 'UTF-8' to array of allowed charsets, to support Postgres

SpikedCola:
After importing the sample SQL into Postgres, I get an error message saying the database needs to be set to utf8, when it already is:

charset

In my case, $resourceConfig['charset'] === 'UTF-8'. By adding this to the list of allowed charsets, the error message disappears.


SpikedCola:
also looks like this possibly fixes #1535

Thomas-Gelf:
Merged, thank you!

Issue #1819, opened 2019-04-04
ImportsourceCommand does not apply modifiers on fetch


Thomas-Gelf:
## Expected Behavior
`icingacli director importsource fetch` output should match imported data

## Current Behavior
It doesn't, configured property modifiers are not applied

PR #1936, opened 2019-09-06
CLI: Apply modifiers in 'importsource fetch'

lippserd:
refs #1819


Thomas-Gelf:
Good catch, thank you!

PR #1962, opened 2019-09-26
Updated Windows Agent installer script

LordHepipud:
Updated Windows Agent installer script to the latest available version


Thomas-Gelf:
Thank you!

PR #2008, opened 2019-11-02
Fixing linux install script version check

julian-weinert:
It seems as if the version output of the Icinga executable changed.

The current version check searches for "version" whereas Icinga 2.1.1-1 outputs the word capitalized.

I added the `-i` flag to grep to ignore the case there instead of complicating the rexex.


Thomas-Gelf:
Thanks!

PR #2037, opened 2019-11-28
use sprintf() for InvalidArgumentException

mantis:


Thomas-Gelf:
Thanks!

mantis:
@Thomas-Gelf https://github.com/Icinga/icingaweb2-module-director/search?q=throw+new+DuplicateKeyException&type=Code - can you scan the code base at some point - i same bug is repeated else where. I just hit this particular one, so edited in web ui.

Thomas-Gelf:
DuplicateKeyException and others that inherit IcingaException should be fine, we implemented variadic args / sprintf in that base class. Errors have been introduced where such Exceptions have been transformed into "native" PHP Exceptions, like InvalidArgumentException.

PR #2054, opened 2020-01-08
Add php-mbstring as requirement

JonasVerhofste:
6ca14a58b0888413a98ac0754a99524004df1c3a started using mbstring, so it needs to be specified as a requirement

### DIFF
--- a/doc/02-Installation.md
+++ b/doc/02-Installation.md
@@ -30,6 +30,7 @@ Requirements
 * php-pcntl (might already be built into your PHP binary)
 * php-posix (on RHEL/CentOS this is php-process, or rh-php7x-php-process)
 * php-sockets (might already be built into your PHP binary)
+* php-mbstring
 
 Database
 --------


Thomas-Gelf:
Thanks @JonasVerhofste! You're right, mbstring is a requirement. It hasn't been added to the list, as Icinga Web 2 already required mbstring at the time being. Still, it wouldn't hurt to list ALL requirements - but then I'd prefer to see a complete list (json, mysql...). Other opinions?

JonasVerhofste:
@Thomas-Gelf Ah, completely overlooked it on the web 2 docs. We can make a list of all of them, or we can make a link to the web 2 docs stating we expect those dependencies to be there already too (and maybe make a list of shared dependencies?)

Thomas-Gelf:
As of this writing, Director has direct code references to the following extensions:

* curl
* iconv
* json
* mbstring
* pcntl
* posix
* sockets

This list does not include indirect dependencies, with pdo* being the most essential and prominent one.

JonasVerhofste:
@Thomas-Gelf how would you propose we word the indirect requirements? Because the pdo-ones probably depend on what DB you're using?

PR #2080, opened 2020-02-11
Add Italian translations

ValeDaRold:
Add files containing Italian traslations for Director module. Added both `.po` and compiled file `.mo`


Thomas-Gelf:
Merged, thank you!!

nilmerg:
@Thomas-Gelf Please transfer changes to the director locales to https://github.com/Icinga/L10n and don't just merge them here. The module locales should only serve as fallback for < Web 2.8.0 users. All others already got icinga-l10n installed and should get theirs from it instead. Though, these changes here didn't made it there and neither did those from a2757f13768051b811d0a6601f98850a4a7c5ebc. I'll do that now, but next time please notify me or tell contributors to.

PR #2085, opened 2020-02-14
Fix macros with subproperties on service-apply-rule (#2083)

### LINKED ISSUE #2083: Using macro sub-properties not working in Service Apply Rules
## Expected Behavior
A macro like `$config.myproperty$` should be evaluated as macro
image

## Current Behavior
A macro like `$config.myproperty$` is evaluated as a string, currently only `$config$` is treated as a macro
image

## Possible Solution
The macro-whitelist could be implemented as a regex-pattern instead of plain string. This way instead of whitelisting `array(['config'])` it could be whitelisted as something like `array(['^config[.$]'])` to also capure all string starting with `config.`

1. Create a Host with a custom variable containing an array of objects
image
2. Create a Service-Apply-Rule for this array
3. Try to use a macro for the property of the array-element (e.g: `$config.host$`)
4. Check the preview, `config.host` is surrounden by `"`

## Your Environment

* Director version (System - About): 1.7.2 (master Feb 12, 2020)
* Operating System and version: Ubuntu 18.04.4 LTS (Bionic Beaver)
* Webserver, PHP versions: apache/2.4.41

semmasu:
### Fixes #2083 

In Service-Apply-Rules macros like `$config.property$` should be evaluated, currently the only macro that gets evaluated is `$config$`

### DIFF
--- a/library/Director/CustomVariable/CustomVariableString.php
+++ b/library/Director/CustomVariable/CustomVariableString.php
@@ -46,7 +46,7 @@ class CustomVariableString extends CustomVariable
     public function toConfigString($renderExpressions = false)
     {
         if ($renderExpressions) {
-            return c::renderStringWithVariables($this->getValue(), ['config']);
+            return c::renderStringWithVariables($this->getValue(), ['^config[.$]']);
         } else {
             return c::renderString($this->getValue());
         }
--- a/library/Director/IcingaConfig/IcingaConfigHelper.php
+++ b/library/Director/IcingaConfig/IcingaConfigHelper.php
@@ -385,7 +385,7 @@ class IcingaConfigHelper
                         // We got a macro
                         $macroName = substr($string, $start + 1, $i - $start - 1);
                         if (static::isValidMacroName($macroName)) {
-                            if ($whiteList === null || in_array($macroName, $whiteList)) {
+                            if ($whiteList === null || preg_match('/(' . implode(')|(', $whiteList) . ')/', $macroName)) {
                                 if ($start > $offset) {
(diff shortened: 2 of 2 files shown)


Yoda-BZH:
I also tried to come up with a solution for this problem.

My solution was:

```patch
diff --git a/library/Director/IcingaConfig/IcingaConfigHelper.php b/library/Director/IcingaConfig/IcingaConfigHelper.php
index ed9f6af7..c8517e64 100644
--- a/library/Director/IcingaConfig/IcingaConfigHelper.php
+++ b/library/Director/IcingaConfig/IcingaConfigHelper.php
@@ -366,6 +366,14 @@ class IcingaConfigHelper
             && ! preg_match('/\.$/', $name);
     }
 
+
+    public static function isMacroDictionnary($macroName, $whiteList = null)
+    {
+        $tokens = explode('.', $macroName);
+
+        return $whiteList === null || in_array($tokens[0], $whiteList);
+    }
+
     public static function renderStringWithVariables($string, array $whiteList = null)
     {
         $len = strlen($string);
@@ -385,7 +393,7 @@ class IcingaConfigHelper
                         // We got a macro
                         $macroName = substr($string, $start + 1, $i - $start - 1);
                         if (static::isValidMacroName($macroName)) {
-                            if ($whiteList === null || in_array($macroName, $whiteList)) {
+                            if ($whiteList === null || in_array($macroName, $whiteList) || static::isMacroDictionnary($macroName, $whiteList)) {
                                 if ($start > $offset) {
                                     $parts[] = static::renderString(
                                         substr($string, $offset, $start - $offset)
```

PR #2095, opened 2020-03-05
Add wildcard support for contains and render with match()

lazyfrosch:
This used MatchAny explicitly, since we check against an array here.

When a user specifies an filter like this in Director:

    host.groups contains "customer*"

The Director should render it as:

    assign where match("customer*", host.groups, MatchAny)

For backwards compatibility I wouldn't change wildcards with `=`, but for contains this new behavior here should be expected.

ref/NC/620757

### DIFF
--- a/library/Director/IcingaConfig/AssignRenderer.php
+++ b/library/Director/IcingaConfig/AssignRenderer.php
@@ -112,11 +112,24 @@ class AssignRenderer
 
     protected function renderContains(FilterExpression $filter)
     {
-        return sprintf(
-            '%s in %s',
-            $this->renderExpressionValue(json_decode($filter->getColumn())),
-            $filter->getExpression()
-        );
+        // Note: expression and column is flipped in this mode
+        $value = json_decode($filter->getColumn());
+        $expression = $this->renderExpressionValue($value);
+        $column = $filter->getExpression();
+
+        if (strpos($value, '*') !== false) {
+            return sprintf(
+                'match(%s, %s, MatchAny)',
+                $expression,
+                $column
+            );
+        } else {
+            return sprintf(
+                '%s in %s',
+                $expression,
+                $column
+            );
+        }
     }
 
     protected function renderFilterExpression(FilterExpression $filter)
--- a/test/php/library/Director/IcingaConfig/AssignRendererTest.php
+++ b/test/php/library/Director/IcingaConfig/AssignRendererTest.php
@@ -93,6 +93,15 @@ class AssignRendererTest extends BaseTestCase
             $expected,
             $this->renderer($string)->renderAssign()
         );
+
+        $string = json_encode('member*') . '=host.vars.some_array';
+
+        $expected = 'assign where match("member*", host.vars.some_array, MatchAny)';
+
+        $this->assertEquals(
+            $expected,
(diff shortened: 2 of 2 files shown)


Thomas-Gelf:
@lazyfrosch: thanks, this would make a nice feature. Did you test whether our (Web) Filters are able to calculate the same result? There must be a discrepancy either before or after this modification I guess

lazyfrosch:
@Thomas-Gelf How would this be related to the web filters? The AssignRenderer is already quite specific to Director. I've only added the recognition of wildcards with the `contains` logic

Please see the updated unit test.

Thomas-Gelf:
@lazyfrosch: relation to web filters -> we cannot deploy Filters we're unable to evaluate by ourselves

lazyfrosch:
I still don't get the point, sorry.

Thomas-Gelf:
`ObjectApplyMatches` must provide the same result as Icinga 2. Either it currently behaves differently than the core, or it will do so after this patch

log1-c:
Any update on this?
We just stumbled upon this, as we have hosts with an array variable filled with tags.
Now having an apply rule like `host.vars.tags contains "tag_thistag"` works without problems, bit `host.vars.tags contains "tag_handle_*"` does not.

Is this to be expected to work any time soon?

khmuth:
Can we expect this in release 1.9?
Because of the missing possibility to add a "MatchAny" we have to maintain .conf files in addition to the director configuration.

sciolto:
Hi @Thomas-Gelf @lazyfrosch  is there any chance this will be pulled and put in a release ? 
We have a big customer which we need to patch at every update.

lazyfrosch:
I haven't been involved with Icinga for a while now, hopefully this will get merged some day...

sciolto:
@Thomas-Gelf have you had the chance to review this PR ? 
thanks
Oreste

nilmerg:
ref/IP/52313

sciolto:
@Thomas-Gelf @nilmerg  any news about the inclusion of this PR in head ?

nilmerg:
It's scheduled for the next major. If that's being worked on, this gets reviewed as well.

sciolto:
Hi @nilmerg any chance this makes into a release ? 
Thanks

lippserd (approved):
As per the discussion between Thomas and Markus when this PR was opened, the main concern was whether the new wildcard support would maintain evaluation parity between Icinga Web's (+ Director) filter syntax and Icinga 2.

In fact, this PR would **fix parity**, as wildcard support for arrays is already given for Icinga Web's filters and Director's object apply:

```php
$object1 = (object) [
    'object_name' => 'object1',
    'vars.some_array' => ['member1', 'member2']
];
$object2 = (object) [
    'object_name' => 'object2',
    'vars.some_array' => ['foo', 'bar']
];
$filterString = json_encode('member*') . '=host.vars.some_array';
$filter = Filter::fromQueryString($filterString);
HostApplyMatches::fixFilterColumns($filter);
$helper = new AssignFilterHelper($filter);
$helper->matches($object1); // true
$helper->matches($object2)); // false

lippserd:
@raviks789 Please rebase and merge.

PR #2101, opened 2020-03-20
Fix host templates not usable with hostgroup restrictions

### LINKED ISSUE #2020: Host Templates are not usable by user having a role with director/filter/hostgroups assigned
## Expected Behavior
I'd like to have different admin users having each a set of servers. They should be able to create new hosts themselves based on templates created by an admin user. I, as admin, have created a host template for each the groups, every host templates having a different HostGroup, i.e:

```
template Host "MOLBIO -Standard Linux Server" {
    import "Standard Linux Server"

    groups = [ "molbio" ]
    vars.department = "molbio"
}
```
I have then created a restricted admin role which has everything enabled but with a restriction on director/filter/hostgroups, i.e. `hostgroup_name=molbio`

## Current Behavior
A user with that restricted role sees host templates but cannot use them, nor create a new template based on them. Moreover, the user cannot create a HostTemplate for himself, but that's better like this.

Whatever the user does with host templates the following message is displayed:

> Unable to store a host with the given properties because of insufficient permissions (IcingaHostForm.php:363)

If the user selects a template and click on the "Host" tab, the following is displayed

> No such object available
> 
> 0 /usr/share/icingaweb2/modules/director/library/Director/Web/Controller/ObjectController.php(441): Icinga\Module\Director\Web\Controller\ObjectController->loadObject()
> 1 /usr/share/icingaweb2/modules/director/library/Director/Web/Controller/ObjectController.php(75): Icinga\Module\Director\Web\Controller\ObjectController->eventuallyLoadObject()
> 2 /usr/share/php/Icinga/Web/Controller/ActionController.php(152): Icinga\Module\Director\Web\Controller\ObjectController->init()
> 3 /usr/share/php/Icinga/Web/Controller/Dispatcher.php(59): Icinga\Web\Controller\ActionController->__construct(Object(Icinga\Web\Request), Object(Icinga\Web\Response), Array)
> 4 /usr/share/icingaweb2/library/vendor/Zend/Controller/Front.php(937): Icinga\Web\Controller\Dispatcher->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
> 5 /usr/share/php/Icinga/Application/Web.php(300): Zend_Controller_Front->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
> 6 /usr/share/php/Icinga/Application/webrouter.php(99): Icinga\Application\Web->dispatch()
> 7 /usr/share/icingaweb2/public/index.php(4): require_once(String)
> 8 {main}

1. Create non-admin role
2. Grant permission everywhere
2. Add a filter to director/filter/hostgroups

## Your Environment

* Director version (System - About): 1.7.2
* Operating System and version: Ubuntu  18.04.3 LTS (Bionic Beaver)
* Webserver, PHP versions: Apache/2.4.29 (Ubuntu), PHP 7.2.24-0ubuntu0.18.04.1

lippserd:
Before, for new hosts the code checked whether the host is part of any
restricted hostgroup only if the host's groups got modified directly.
I honestly don't know how that should be possible, since one can't add
host groups directly if restricted.
Now, for new hosts the code checks whether the resolved host groups
match at least one of the restricted host groups.

fixes #2020

### DIFF
--- a/library/Director/Restriction/HostgroupRestriction.php
+++ b/library/Director/Restriction/HostgroupRestriction.php
@@ -57,11 +57,9 @@ class HostgroupRestriction extends ObjectRestriction
         }
 
         if (! $host->hasBeenLoadedFromDb()) {
-            if ($host->hasModifiedGroups()) {
-                foreach ($this->listRestrictedHostgroups() as $group) {
-                    if ($host->hasGroup($group)) {
-                        return true;
-                    }
+            foreach ($this->listRestrictedHostgroups() as $group) {
+                if ($host->hasGroup($group)) {
+                    return true;
                 }
             }
 


lippserd:
Hi @Thomas-Gelf,

It is currently not possible for restricted users to create hosts from allowed templates.

With this PR it's fixed 😆 

There's another issue that restricted users may choose from every created template - even if the template's groups do not match the restricted groups. Which will then lead into an exception.
Should I open a dedicated issue for this or is this on your roadmap anyway?

All the best,
Eric

Thomas-Gelf:
@lippserd: if I can recall it correctly the initial request for this feature has been satisfied with "allow to modify hosts and their services, especially to override thresholds on services deriving from Apply Rules. Hosts have been created in an automated way, so this hasn't been an issue. And who really needed to create a new one was able to clone and modify an existing one - that worked fine.

Your patch is welcome, there is no reason why manually creating a new host (fitting the rules) shouldn't be allowed. Just, and that's where it becomes tricky: if you do not know the configured rules you might have difficulties to choose the correct template. Especially when you're facing a lot of them.

Showing only templates matching your Hostgroup Restriction will not work, as you might want to combine multiple templates for different reasons. You cannot expect the "Run faster Checks" Template to be there as a dedicated instance for each Hostgroup.

Cheers,
Thomas

lippserd:
Thanks for the response and I got your point. But we could list all templates that don't define a hostgroup and all templates that match the restriction, right?

Thomas-Gelf:
Sure, that's one black magic variant that could mostly work :D However, why shouldn't you be allowed to (additionally) choose the "LAN Servers" Template (with a related Hostgroup) when being restricted to Hosts in the "Linux Servers" Hostgroup? It's quite tricky to get this right.

Alternatively I'd suggest to just teach "your" restricted users to clone existing hosts in case they need to create new ones. As they usually control a subset of very similar Hosts with not so many variants, that's much easier to grasp.

At least in the meantime, because there is more to come. There is an extended variant of the "Template Choices" feature in the make. It will allow admins to enforce a specific "root" template, with all the others being "Choices". Your users will then create their hosts in a wizard-like way, and the word "Template" will not even show up in their frontend.

lippserd:
> Sure, that's one black magic variant that could mostly work :D However, why shouldn't you be allowed to (additionally) choose the "LAN Servers" Template (with a related Hostgroup) when being restricted to Hosts in the "Linux Servers" Hostgroup? It's quite tricky to get this right.

Yeah, makes sense. That would a require at least a 2-step form. So the upcoming choices feature sounds really promising. Anyway, enjoy your wine as I do mine 😆 and have a good night 🛌

Thomas-Gelf:
2-step: it depends on how you're going to configure it. Might be single- or multi-step - that's completely up to you. Choices are already there, you'll just get a very few more related toggles.

For example, there is the `use_as_default` property labelled **Main Choice** with the following description:

> You might want to specify a single Choice being the required main Choice for all hosts. When doing so, the main "imports" element will no longer be available for non-admin users.

At least that's what the branch looked like two years ago. As of today, I'd probably opt for allowing you to have "multiple root templates". If a user has access to only one of them, he'll have a single-step form. Otherwise, he has of course to first opt for a specific starting point. Apart from your root Choice, there might be many more of them - depending on your permissions and related template/choice dependencies. We'll see. Have to prepare some other basics first before I'll reanimate that ancient branch.

I'm still busy with the tiramisu over here ;-) Cheers :wine_glass:

PR #2132, opened 2020-06-05
ImportSourceRestApi: Add header and deeper extract_property

lazyfrosch:
Allows a user to add additional headers, e.g. by setting a specific `Accept` or any authentication header.

Also `extract_property` now has a logic for deeper keys like "result.objects", "key.deeper_key.very_deep"


Thomas-Gelf:
Thanks!

PR #2161, opened 2020-07-14
Make 50:50 layout working with the flexbox layout

### LINKED ISSUE #2065: Column with issues with .container { width != 0 }
Director is one of the modules using a custom column width. It sticks with 50/50 instead of the default 33/66 on wider screens. This works fine up to v2.7.3, but leads to the following result with the current master:

web2_master-cols

To reproduce this go to any host in the monitoring module an click on "Modify".

nilmerg:
fixes #2065

### DIFF
--- a/public/css/module.less
+++ b/public/css/module.less
@@ -964,14 +964,21 @@ form.db-selector {
   }
 }
 
-#layout.twocols &#col1 {
+// Adjustments for the legacy layout to keep backwards compatibility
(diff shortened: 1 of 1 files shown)


nilmerg:
@Thomas-Gelf Any objections against that I merge this? If not and you read this in the next 3 days, you can also merge it yourself right away. Otherwise I do.

PR #2180, opened 2020-08-31
Add zones to Hostgroups and Servicegroups

eurotux-tec:
In the scenario of our company there is an icinga master node and several satellites, each one is responsible for monitoring the infrastructure of a specific customer.

Currently, all hostgroups / servicegroups are replicated to all satellites, so all customers are aware of the other hostgroups / servigroups, which should be avoided.

This PR contains changes to the PHP code and DB schema to allow a zone to be assigned to a hostgroup / servicegroup, so that the hostgroup / servicegroup configuration file is only sent to the corresponding zone.

Do you think that this makes sense or can we do it in a different way? If you agree, please merge this. If you don't, please let me know how can I improve the patch so that is gets merged.

### DIFF
--- a/application/forms/IcingaGroupForm.php
+++ b/application/forms/IcingaGroupForm.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Icinga\Module\Director\Forms;
+
+use Icinga\Module\Director\Web\Form\DirectorObjectForm;
+
+class IcingaGroupForm extends DirectorObjectForm
+{
+    /**
+     * @return $this
+     * @throws \Zend_Form_Exception
+     */
+    protected function addZoneElements()
+    {
+        $this->addZoneElement(true);
+        $this->addDisplayGroup(['zone_id'], 'clustering', [
+            'decorators' => [
+                'FormElements',
+                ['HtmlTag', ['tag' => 'dl']],
+                'Fieldset',
+            ],
+            'order'  => self::GROUP_ORDER_CLUSTERING,
+            'legend' => $this->translate('Zone settings')
+        ]);
+
+        return $this;
+    }
+}
--- a/application/forms/IcingaHostGroupForm.php
+++ b/application/forms/IcingaHostGroupForm.php
@@ -2,10 +2,11 @@
 
 namespace Icinga\Module\Director\Forms;
 
-use Icinga\Module\Director\Web\Form\DirectorObjectForm;
-
-class IcingaHostGroupForm extends DirectorObjectForm
+class IcingaHostGroupForm extends IcingaGroupForm
 {
+    /**
+     * @throws \Zend_Form_Exception
+     */
     public function setup()
     {
         $this->addHidden('object_type', 'object');
@@ -20,21 +21,6 @@ class IcingaHostGroupForm extends DirectorObjectForm
              ->addAssignmentElements()
              ->addZoneElements()
              ->setButtons();
-    }
-
-    protected function addZoneElements()
-    {
-        $this->addZoneElement(true);
-        $this->addDisplayGroup(['zone_id'], 'clustering', [
-            'decorators' => [
-                'FormElements',
-                ['HtmlTag', ['tag' => 'dl']],
-                'Fieldset',
-            ],
-            'order' => 80,
-            'legend' => $this->translate('Zone settings'),
-        ]);
-        return $this;
     }
 
     protected function addAssignmentElements()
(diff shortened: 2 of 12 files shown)


araujorm:
Hello.

Sorry for having had this stalled for so long, but I made the corrections that were asked, leaving only what is *indeed* relevant, adapted to be on par with the current master version (uuid property added, database version and so on), and refreshed the branch used for this pull request.

Can you review it, and if good, ponder about merging it? Anything that might be missing please tell.

Thanks for your work.

araujorm:
> Any objection against unifying Service/Host/UserGroup code for Objects and Forms, at least as far as possible? I kind of dislike the duplicated code. Parts of it already used to be there, but now it get's even more.

@Thomas-Gelf Had missed this part. Meanwhile, took a look a the code and tried to unify it.

We added a new commit to the pull request (in order to have it clear what was made to add support for zones on hostgroups and service groups, and what was made to unify the code afterwards). 

For objects, moved some common parts to IcingaObjectGroup, where zone_id is now in defaultProperties and thus those don't need to be overriden in IcingaHostGroup and IcingaServiceGroup anymore. Similar for relations property, which don't need to be specified in descendant classes anymore.

For forms, created a intermediate class IcingaGroupForm (analog to the already existing IcingaObjectGroup for objects) which extends DirectorObjectForm, and made IcingaHostGroupForm,  IcingaServiceGroupForm and IcingaUserGroupForm extend this new class instead. Moved the addZoneElements method to this new class (although it's not used by IcingaUserGroupForm), thus removing it from the descendant classes that had it repeated.

Was this what you had in mind, @Thomas-Gelf ? If you think this should be on a separate pull request, we can remove this new commit and make another pull request later after the zone support for hostgroups and servicegroups is merged. Please let us know, will await your feedback.

araujorm:
Made a correction on the last commit message, because the addZoneElements method **is** used by all the descendant classes, including IcingaUserGroupForm.

Hope this is what is intended, best regards.

eurotux-tec:
Hello.

We've refreshed the proposed commits so that they merge with the current master.

Could we have any feedback on whether it's acceptable, or if there is any other approach you'd prefer instead?

araujorm:
Good afternoon. Branch of the merge request was rebased so that the MR is valid for latest release and master branch.

eurotux-tec:
Hello,

We have rebased this pull request onto the current master and would appreciate a fresh review.

Since the branch was last updated, master has evolved in this area, so we reconciled our changes with it rather than reintroducing older assumptions. A summary of what we verified had changed and how we adapted:

Objects (library/Director/Objects)
- Upstream now keeps `assign_filter` in the group `defaultProperties`, and has added the membership-resolver plumbing (`beforeStore()`/`notifyResolvers()`/`getMemberShipResolver()`) plus `supportsAssignments()`. We kept all of that.
- The unification of the group objects is largely aligned with master already (the concrete classes extend `IcingaObjectGroup`). We moved `zone_id` and the `zone` relation into `IcingaObjectGroup`, and left `prefersGlobalZone()` returning `true` in the base (so Host- and ServiceGroups default to the global zone, as suggested previously). `IcingaUserGroup` keeps `prefersGlobalZone()` returning `false`, since user objects normally live in the master zone.

Forms (application/forms)
- We reintroduced the intermediate `IcingaGroupForm` (extending `DirectorObjectForm`) and made `IcingaHostGroupForm`, `IcingaServiceGroupForm` and `IcingaUserGroupForm` extend it, moving the shared `addZoneElements()` there.
- In the meantime, master added an assignment filter and a delete redirect to `IcingaUserGroupForm` (`addAssignmentElements()` and `deleteObject()`). We preserved both. As with the other two group forms, `addAssignmentElements()` stays in the subclass because its `suggestionContext` differs per type.

Schema
- The zones migration was renumbered to the next free version on the current master (193 for both mysql and pgsql), and the full schema version was bumped accordingly.
- The PostgreSQL migration and full schema had carried MySQL-specific syntax (`INT(10) UNSIGNED`, `FOREIGN KEY zone (...)`), which is invalid on PostgreSQL. This has been corrected to valid PostgreSQL (`integer`, `FOREIGN KEY (zone_id)`, plus the matching `CREATE INDEX ..._zone`), consistent with how the other zone-aware tables are defined.

The second commit (unifying the Service/Host/UserGroup code) was originally added at a maintainer's request. If you would now prefer to keep this PR focused strictly on adding zone support to Host- and ServiceGroups, we are happy to drop that commit and open the unification as a separate pull request. Please let us know which you prefer.

Thanks for your work.

PR #2181, opened 2020-08-31
Add deploy option for Sync jobs

eurotux-tec:
In our company, we use Icinga Director automations to create the customer infrastructure when it's in the cloud (AWS).

Currently, there is an import source responsible for collecting information about the servers to be monitored and a sync rule in charge of creating the objects (hosts).

Since it's an infrastructure in the cloud it's highly dynamic, so an automation is necessary to keep the monitoring up to date.
For this, we created three jobs to execute periodically (5 minutes) to run the import source, the sync rule and the configuration deployment.
Considering the current approach, configurations are deployed every 5 minutes, regardless of whether there are changes or not. In our scenario where there are many satellites and each deployment implies sending the configuration to all satellites, the impact is significant.

In this PR, a change to Sync Job is proposed so that it is responsible not only for creating the configuration but also for deploying it (when necessary).

Do you think that this makes sense or can we do it in a different way? If you agree, please merge this. If you don't, please let me know how can I improve the patch so that is gets merged.

### DIFF
--- a/library/Director/Job/SyncJob.php
+++ b/library/Director/Job/SyncJob.php
@@ -6,6 +6,8 @@ use Icinga\Module\Director\Hook\JobHook;
 use Icinga\Module\Director\Web\Form\DirectorObjectForm;
 use Icinga\Module\Director\Web\Form\QuickForm;
 use Icinga\Module\Director\Objects\SyncRule;
+use Icinga\Module\Director\Deployment\ConditionalConfigRenderer;
+use Icinga\Module\Director\Deployment\ConditionalDeployment;
 
 class SyncJob extends JobHook
 {
@@ -14,17 +16,38 @@ class SyncJob extends JobHook
     /**
      * @throws \Icinga\Exception\NotFoundError
      * @throws \Icinga\Module\Director\Exception\DuplicateKeyException
+     * @throws \Icinga\Exception\IcingaException
      */
     public function run()
     {
         $db = $this->db();
         $id = $this->getSetting('rule_id');
+
+        $shouldDeploy = false;
+        switch ($this->getSetting('deploy', 'never')) {
+            case 'side_effect_free':
+                $shouldDeploy = $db->countActivitiesSinceLastDeployedConfig() == 0;
+                break;
+            case 'always':
+                $shouldDeploy = true;
+                break;
+        }
+
+        $madeChanges = false;
         if ($id === '__ALL__') {
             foreach (SyncRule::loadAll($db) as $rule) {
-                $this->runForRule($rule);
+                if ($this->runForRule($rule)) {
+                    $madeChanges = true;
+                }
             }
         } else {
-            $this->runForRule(SyncRule::loadWithAutoIncId((int) $id, $db));
+            $madeChanges = $this->runForRule(SyncRule::loadWithAutoIncId((int) $id, $db));
+        }
+
+        if ($madeChanges && $shouldDeploy) {
+            $deployer = new ConditionalDeployment($db);
+            $renderer = new ConditionalConfigRenderer($db);
+            $deployer->deploy($renderer->getConfig());
         }
     }
 
@@ -35,7 +58,8 @@ class SyncJob extends JobHook
     public function exportSettings()
     {
         $settings = [
-            'apply_changes' => $this->getSetting('apply_changes') === 'y'
+            'apply_changes' => $this->getSetting('apply_changes') === 'y',
+            'deploy' => in_array($this->getSetting('deploy'), array('side_effect_free','always'), true)
         ];
         $id = $this->getSetting('rule_id');
         if ($id !== '__ALL__') {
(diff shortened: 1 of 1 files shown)


Al2Klimov:
Hello @eurotux-tec and thank you for contributing!

Imagine you've configured that job, update Director and suddenly the job does much more than you expect. Would you like this kind of behavior in your production? (I'd flip my table.)

> Considering the current approach, configurations are deployed every 5 minutes, regardless of whether there are changes or not.

@Thomas-Gelf Can you confirm this?

https://github.com/Icinga/icingaweb2-module-director/blob/d804448d79531e6b4349d69b45b4cba679ce8640/library/Director/Job/ConfigJob.php#L31-L33

Best,
AK

Thomas-Gelf:
@Al2Klimov: you can have an automated Config Deploy Job scheduled to run every minute or even faster:

* it will not deploy unless the generated configuration differs from the last deployed configuration
* it will not even spend time and try to render a new configuration unless
   * there are un-deployed activities in the Activity Log
   * you enabled the flag "Force Rendering"

If you then also define a grace period you can for example make sure that it deploys at most every X minutes, even if there are changes every few seconds. And when you select a Time Period you can for example make sure that automated Deployments occur only during office hours.

Well - and that's it. Works even in large environments with a massive amount of changes and lots of Import Sources and Sync Rules.

@eurotux-tec: thanks for your effort, but as @Al2Klimov already pointed out this change would be pretty unexpected. Quite some environments have only automated Import and Sync Jobs, but prefer to deploy manually as they do not trust their various config sources.

You wrote:

> Considering the current approach, configurations are deployed every 5 minutes, regardless of whether there are changes or not.

Are you sure about this? This would be a severe bug.

araujorm:
Hello.

There are situations when having a deploy forced every 5 minutes is undesirable, because then you can't be adding new stuff without it being deployed without wanting yet. So I prefer the alternative proposed by this patch, even knowing I can still have a deploy being made once in a while when I'm editing other stuff, but at least it's not always, and it will be because my very important sync job had to do it.

So, maybe not by default, since not everyone wants to have automated deploys, but to have this as an option is a must on certain environments. If it can be made optional, would anyone disagree on it being merged?

eurotux-tec:
Hello.

Patch has been reworked to not affect the default behaviour. Instead, a new option "Deploy" has been added to the Sync Job config. That option is only visible/affects if "Apply changes" is "Yes".

By default, the "Deploy" option is "No". Other possible values are:
- Yes: if there are changes made by this sync job, a deploy will be made at the end, but only if there were no other pending changes before (so that an automated job doesn't deploy changes being made by humans at that same time)
- Yes (forced): a deploy will always be made (provided there are changes made by this sync job) even if there were other pending changes beforehand.

With these changes, and with the new feature being completely optional, is this pull request now acceptable?

Best regards.

Thomas-Gelf:
Some thoughts:

We recently implemented Configuration Branches for the Icinga Director. The code is in place and working, schema changes have already been applied. The related GUI controls are (intentionally) still missing.

We're now able to apply changes to Configuration Branches, which are safe different realities - similar to branches in your VCS. Sync runs are part of this whole picture. Plans are to offer a "Sync to a temporary Branch" functionality. This will allow one to navigate all changes applied by a Sync rule directly in the Director UI, and to then merge or reject those changes at will, and all at once.

Also related: we implemented Push-only Import Sources, where instead of pulling data, you ship it to the Import Source (HTTP POST or similar). This is somewhat hidden, and leveraged only by some (or at least one) add-on modules. There we also trigger related Sync Rules in case the imported data changed.

This will probably ask for some additional flags on the Import and Sync objects themselves. Wanted to mention this, as you might have to expect some delay for this feature. Depending on where they move, we might prefer to shift your setting from the Job to the Sync rule itself.

Cheers,
Thomas

araujorm:
Hello.

Remade the proposed patch as follows:
- updated to merge with current master
- usage of camelCase for variable names and renamed introduced variables as such
- explicitly name the expected default value in the code when using getSetting
- reformulated some execution logic in order to only see if deploys should be made *after* knowing changes were made, and there use a switch statement for better readability
- renamed the deploy settings to side_effect_free ("Yes (side effect free)"), always ("Yes (always)") and never ("Never")
- trigger the deploy using ConditionalDeployment and ConditionalConfigRenderer objects, as implemented by ConfigJob in more recent code, thus avoiding the rest of the problems previously pointed bt @Thomas-Gelf

@Thomas-Gelf Can you see if it's good? Many thanks.

araujorm:
Sorry, last version missed importing the necessary namespaces for ConditionalConfigRenderer and ConditionalDeployment, and needed to remove some others used in the first version of this patch. Fixed that in latest update to the branch.

araujorm:
Hello.

We've refreshed the proposed commit so that it merges with the current master.

Could have any feedback on whether it's acceptable, or if there is any other approach you'd prefer instead?

araujorm:
Good afternoon. Branch of the merge request was rebased so that the MR is valid for latest release and master branch.

eurotux-tec:
Hello,

This pull request has been rebased onto the current master. It applied cleanly, with no conflicts, and the functionality is unchanged: it adds a deploy option to Sync jobs (`SyncJob.php`), so that a successful sync run can optionally trigger a deployment.

We would appreciate a fresh review and, if it looks good, a merge. Please let us know if anything is missing.

Thanks for your work.

PR #2186, opened 2020-09-24
Daemon: start after database

dgoetz:
This changes the daemon to start after network being really up and to wait for a database so it does not start with an error.

### DIFF
--- a/contrib/systemd/icinga-director.service
+++ b/contrib/systemd/icinga-director.service
@@ -1,7 +1,8 @@
 [Unit]
 Description=Icinga Director - Monitoring Configuration
 Documentation=https://icinga.com/docs/director/latest/
-Wants=network.target
+Wants=network-online.target
+After=mariadb.service mysqld.service postgresql.service
 
 [Service]
 EnvironmentFile=-/etc/default/icinga-director


Thomas-Gelf:
Hi @dgoetz,

thank you for this contribution, but I'm absolutely not a friend of doing this. The daemon should always start immediately and keep running. Systemd shouldn't tear it down when someone restarts the DB. Personally I'm never running the DB on the very same system/container, I hate systems misbehaving because of paranoid dependencies in unit files.

Also, I do not care whether the DB is ready or not, the Daemon MUST start. As soon as it is able to reach the DB it will carry on it's work. Initially it will try to reconnect in very short intervals, after some time it slows down it's reconnection attempts - but it never stops.

Cheers,
Thomas

Al2Klimov:
> systems misbehaving because of paranoid dependencies in unit files

Just out of curiosity: Please could you give a real world example?

Thomas-Gelf:
> > systems misbehaving because of paranoid dependencies in unit files
> 
> Just out of curiosity: Please could you give a real world example?

Even the assumption that the DB is running on the same system is wrong. Instead, there might for whatever reason be a completely unrelated DB-Service being installed and/or running on the same host. Depending on the network-target can result in nice "hickups" on some systems. The daemon is perfectly able to deal with missing/failing DBs and will signal that state. Shutting the daemon down just because someone upgrades (and restarts) the DB leads to useless error messages. Instead of telling the user that the DB went away it will be told that that daemon has gone. Such dependencies have the potential to obscure things and to delay troubleshooting.

These (and more) are just objections for the current state of the software. Upcoming Director versions no longer pick the daemon state from the DB but via Socket. So, the frontend will continue to run even if the daemon and/or the frontend is unable to reach the DB. Currently the user is being told that "the daemon stopped working 5mins ago", in future he will also be able to read "Connection to the DB has been lost 5mins ago, reconnecting..." - like you're currently able to see in the systemctl status.

Closing this, as it would not only potentially cause trouble today, but it would also complete break upcoming features.

PR #2213, opened 2020-10-28
Fixed: GetHostByAddr: Handle Null

Nathaniel-Donahue:
Modifier GetHostByAddr fails for some reason the value is null. Instead of failing this will return null.
Depending on PHP version Lowercase requires PHP-mbstring which was not listed as a dependency.

### DIFF
--- a/library/Director/PropertyModifier/PropertyModifierGetHostByAddr.php
+++ b/library/Director/PropertyModifier/PropertyModifierGetHostByAddr.php
@@ -29,7 +29,10 @@ class PropertyModifierGetHostByAddr extends PropertyModifierHook
 
     public function transform($value)
     {
-        $host = gethostbyaddr($value);
+	if ($value === null) {
+	     return null;
+	}
+	$host = gethostbyaddr($value);
         if ($host === false) {
             switch ($this->getSetting('on_failure')) {
                 case 'null':


doc/02-Installation.md
```diff
@@ -27,6 +27,7 @@ Requirements
   for MySQL we suggest using at least 5.5.3, for PostgreSQL 9.4.
 * PHP (>= 5.6.3). For best performance please consider use 7.x
 * php-curl
+* php-mbstring (might already be built into your PHP binary)
```
Thomas-Gelf, line 30: It's correct that mbstring is required, but it is intentionally not listed as it is already a requirement for Icinga Web 2.

library/Director/PropertyModifier/PropertyModifierGetHostByAddr.php
```diff
@@ -29,7 +29,8 @@ public function getName()
 
     public function transform($value)
     {
-        $host = gethostbyaddr($value);
+        if (is_null($value)) { return null; }    
+	$host = gethostbyaddr($value);
```
Thomas-Gelf, line 33: Please respect [PSR-2](http://www.php-fig.org/psr/psr-2/) - you can use [phpcs](https://github.com/squizlabs/PHP_CodeSniffer) to check your syntax. And please use `$value === null` instead of `is_null`.

Thomas-Gelf (changes requested):
Thanks for your contribution, @Nathaniel-Donahue! I left two comments inline, please address them and then squash/force-push your branch to update this pull request.

Nathaniel-Donahue:
Thanks for the review I adjusted it based on your comment.

Thomas-Gelf (approved):
Great, thank you!

PR #2249, opened 2020-12-08
Updated the doc - PostgreSQL-part

ackbyte:
Corrected the PostgreSQL statements, because of some typos.


Thomas-Gelf:
Thanks for your contribution, @ackbyte! What went wrong with the original code block, didn't that work for you?

ackbyte:
Hi @Thomas-Gelf, there are multiple issues, that getting fixed by this pull request:

1. faulty SQL-Statement syntax, which is not run able because of missing quotation marks
2. mixing multiple execution styles of the statements. One from inside the psql-shell and the other one from the bash-shell

BR
ackbyte

Thomas-Gelf:
Did you try to paste the whole block at once?

ackbyte:
@Thomas-Gelf, no. Because one part of the block is optional... as described in the documentation as an optional part. 
So the whole block has also to be spitted to fit the description.

Thomas-Gelf:
IMHO we should better no longer state that it is "optional" :D Thanks for pointing me to this. Of course if you leave out part of the statement you are required to do so in a correct way. And having separate statements would obviously make it easier to do so. But apart from that, the original block should be correct - isn't it? Or do I miss something?

ackbyte:
That's correct @Thomas-Gelf ;) When pgcrypto gets mandatory part - the origin block is correct and can be used as it is :)

Thomas-Gelf:
Thanks @ackbyte! I'll double-check this and then either merge your pull request or declare the "optional" part being required.

raviks789:
Closing this one out, pgcrypto isn't part of the setup anymore. It got dropped in https://github.com/Icinga/icingaweb2-module-director/commit/b17646534e7ad14db52b2841f43e25c0992cbd51 so this fix doesn't apply to the current docs.

Thanks for the contribution though!

PR #2298, opened 2021-03-10
 Support the rendering of service custom variable in Dependencies form

### LINKED ISSUE #2289: Renders wrong config
## Expected Behavior

Deploying a new cofiguration after an Upgrade led to this error:

```
critical/config: Error: Dependency "....." references a parent host/service which doesn't exist.
Location: in [stage]/zones.d/director-global/dependency_apply.conf: 8:1-8:43
[stage]/zones.d/director-global/dependency_apply.conf(6): }
[stage]/zones.d/director-global/dependency_apply.conf(7): 
[stage]/zones.d/director-global/dependency_apply.conf(8): apply Dependency "ApacheVsVhost" to Service {
                                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[stage]/zones.d/director-global/dependency_apply.conf(9):     import "Dependency Template"
[stage]/zones.d/director-global/dependency_apply.conf(10): 
```

The dependency_apply.conf should look like this. 
```
apply Dependency "TomcatVsAJPPort" to Service {
    import "Dependency Template"
    assign where host.vars.tomcat_ajps && match("AJP Port *", service.name)
    parent_service_name = service.vars.parent_service_name
}
```

## Current Behavior

image

But the config looks like this.

```
apply Dependency "TomcatVsAJPPort" to Service {
    import "Dependency Template"
    assign where host.vars.tomcat_ajps && match("AJP Port *", service.name)
    parent_service_name = "$service.vars.parent_service_name$"
}


```

## Your Environment

* Director version (System - About): Current Master (e2a88861487e9cefad94776fd610b3a9703b9e03)
* Icinga Web 2 version and modules (System - About): 2.8.3
* Icinga 2 version (`icinga2 --version`): 2.12.0-506
* Operating System and version: Centos 7
* Webserver, PHP versions: Apache, PHP 7.3

raviks789:
Earlier the director did not support the rendering of service custom variables for Parent Service in Dependencies form. Here, this issue is fixed.

fixes #2289

### DIFF
--- a/application/forms/IcingaDependencyForm.php
+++ b/application/forms/IcingaDependencyForm.php
@@ -5,6 +5,7 @@ namespace Icinga\Module\Director\Forms;
 use Icinga\Module\Director\Data\Db\DbObject;
 use Icinga\Module\Director\Web\Form\DirectorObjectForm;
 use Icinga\Module\Director\Objects\IcingaDependency;
+use Zend_Validate_Callback;
 
 class IcingaDependencyForm extends DirectorObjectForm
 {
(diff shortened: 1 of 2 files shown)


application/forms/IcingaDependencyForm.php
```diff
@@ -212,11 +212,18 @@ protected function addObjectsElement()
 
         if (!empty($sentParent) || $dependency->isApplyRule()) {
             $parentService = $dependency->get('parent_service');
+            if ($parentService === null) {
+                $parentServiceVar = $dependency->get('parent_service_by_name') ?? '';
+                if (strlen($parentServiceVar) > 0) {
+                    $parentService = '$' . $dependency->get('parent_service_by_name') . '$';
```
nilmerg, line 218: * Drop the null coalesce operation
* Don't use strlen if the actual length is not of interest to you
* Use the variable instead of accessing the property again

application/forms/IcingaDependencyForm.php
```diff
@@ -212,11 +212,18 @@ protected function addObjectsElement()
...
+                }
+            }
             $this->addElement('text', 'parent_service', [
                     'label' => $this->translate('Parent Service'),
                     'description' => $this->translate(
                         'Optional. The parent service. If omitted this dependency'
-                         . ' object is treated as host dependency.'
+                        . ' object is treated as host dependency. You might want to refer'
```
nilmerg, line 225: *refer to*

application/forms/IcingaDependencyForm.php
```diff
@@ -296,14 +303,20 @@ protected function handleProperties(DbObject $object, &$values)
...
         }
 
         parent::handleProperties($object, $values);
     }
 
     protected function isCustomVar($string)
     {
-        return \preg_match('/^\$(?:host)\.vars\..+\$$/', $string);
+        return preg_match('/^\$(?:host|service)\.vars\..+\$$/', $string);
         // Eventually: return \preg_match('/^\$(?:host|service)\.vars\..+\$$/', $string);
```
nilmerg, line 320: drop the comment now

library/Director/Objects/IcingaDependency.php
```diff
@@ -83,6 +83,11 @@ public function parentHostIsVar()
         return $this->get('parent_host_var') !== null;
     }
 
+    protected function isServiceCustomVar($string)
+    {
+        return preg_match('/^(?:service)\.vars\..+$/', $string);
```
nilmerg, line 88: Theres no need for a group to match `service`

library/Director/Objects/IcingaDependency.php
```diff
@@ -594,7 +606,9 @@ protected function getRelatedProperty($key)
         // handle special case for plain string parent service on Dependency
         // Apply rules
         if ($related === null && $key === 'parent_service'
-            && null !== $this->get('parent_service_by_name')
+            && (null !== $this->get('parent_service_by_name')
+                && !$this->isServiceCustomVar($this->get('parent_service_by_name'))
```
nilmerg, line 610: you know the drill: `! ...`

application/forms/IcingaDependencyForm.php
nilmerg: This shouldn't be suggested in case of host dependencies, as for them this is not working. So, only suggest this for service dependencies and add an validator in any case which fails if the user still tries to do this for host dependencies.

nilmerg (changes requested):

application/forms/IcingaDependencyForm.php
```diff
@@ -212,18 +213,46 @@ protected function addObjectsElement()
...
-                         . ' object is treated as host dependency.'
-                    ),
-                    'class' => "autosubmit director-suggest",
-                    'data-suggestion-context' => 'servicenames',
-                    'data-suggestion-for-host' => $sentParent,
-                    'order' => 20,
-                    'value' => $parentService
-                ]);
+                'label'                    => $this->translate('Parent Service'),
+                'description'              => $this->translate($description),
```
nilmerg, line 248: As discussed, this doesn't work. You can only translate literal strings, so either the two parts above individually or both variants on their own in full.

nilmerg (changes requested):

application/forms/IcingaDependencyForm.php
```diff
@@ -212,18 +213,50 @@ protected function addObjectsElement()
...
+            if ($applyTo === 'service') {
+                $description .= $this->translate(
+                    ' You might want to refer to Service Custom Variables via $service.vars.varname$'
+                );
+            }
+
+            $callbackValidator = new Zend_Validate_Callback(function ($value) {
+                $applyTo = $this->getSentValue('apply_to');
+                if ($applyTo === 'service') {
+                    return true;
```
nilmerg, line 239: In case of a service dependency, the user is able to enter `$host.vars.foo`. This may be fine, if it would be rendered as host custom variable, which it currently is not. Check if Icinga is able to resolve a host custom variable in case of a service dependency. If it works, don't render such as strings. If it doesn't, don't accept such as valid values.

application/forms/IcingaDependencyForm.php
```diff
@@ -212,18 +213,50 @@ protected function addObjectsElement()
...
+                }
+
+                return ! $this->isCustomVar($value);
```
nilmerg, line 242: This should complain about service custom variables in case of a host dependency. Such shouldn't be silently accepted and rendered as strings.

application/forms/IcingaDependencyForm.php
```diff
@@ -212,18 +213,50 @@ protected function addObjectsElement()
...
+            });
+
+            $callbackValidator->setMessage(
+                $this->translate('The parent service cannot be a custom variable for host dependency'),
```
nilmerg, line 246: *for **a** host*
or
*for host dependenc**ies***

nilmerg (changes requested):

application/forms/IcingaDependencyForm.php
raviks789: Icinga was able to resolve host custom variables. Hence, both host and service custom variable string patterns are rendered as custom variables.

application/forms/IcingaDependencyForm.php
raviks789: this condition changed to `$applyTo === 'service' && $this->isCustomVar($value)` as we need to check whether the parent service name is a custom variable only for service dependencies.

application/forms/IcingaDependencyForm.php
```diff
@@ -196,34 +197,87 @@ protected function addObjectsElement()
...
+
+            if ($applyTo === 'service') {
+                $parentServiceDescription .= ' ' . $this->translate(
+                    'You might want to refer to Host or Service Custom Variables via $host|service.vars.varname$'
+                );
+            }
+
+            $parentServiceValidator = new Zend_Validate_Callback(function ($value) use ($applyTo) {
+                if ($applyTo === 'host') {
+                    return ! $this->isCustomVar($value);
```
nilmerg, line 260: This must allow host variables. Only service variables cannot be used.

nilmerg (changes requested):

nilmerg (approved):

PR #2350, opened 2021-07-07
Dependency checks dependent on web version

lippserd:


Thomas-Gelf:
### Missing incubator module

This patch removes all checks, which makes error handling worse for the incubator. Without this patch:

no_patch

With this patch:

patch

Tested with disabled incubator module, similar result for no incubator at all. This is easy to fix, I'll skip only ipl and reactbundle dependency checks when there is a v2.9.

### Missing ipl

It's getting even worse for installations where the moved ipl is missing:

ipl-fail

IPL is now a Web 2 requirement, so there isn't much the Director can do about this. However, I would have expected our error handling to be _slightly_ more user-friendly :angry: The Director dependency check page intentionally duplicates code and avoids ipl or incubator components. Reason: it wants to make sure, to get all dependency errors caught. But when our core components fail with fatal errors, there's not much a module can do. Nobody will blame the module, of course. Still, quite annoying.

lippserd:
Icinga Web 2 packages install the IPL and the thirdparty libs. I don't see any reason to check this.
Dependencies are checked when enabling modules.
Though it makes sense to check whether we can catch those errors and indicate that dependencies are missing.
Also a CLI check or something makes sense maybe.
But I highly discourage checking dependencies on every single request.

Thomas-Gelf:
Not everybody uses our packages. Icinga Web 2 fails badly when IPL is missing and doesn't care about thirdparty at all. Web and module dependencies might and will diverge over time, providing meaningful error messages is IMHO essential. I'd love to avoid those checks, and I'm sure that we'll soon be able to obsolete them at framework level.

In the meantime, performance shouldn't be an issue - as this is fast in case all modules are loaded. Been doing this since quite some time now. When dependencies are not satisfied this is slightly slower, but in that case performance doesn't really matter.

I merged a refactored version of those checks, supporting all web versions. It's now also able to complain about outdated libraries. Imagine a new IPL version providing a new widget, in an IPL version required starting with v2.10. Director would badly fail on 2.9 if we wouldn't run those checks. Doing so, users get helpful information and are provided with links pointing to the solution of their problem.

You're right, we should absolutely try to catch those fatal errors and provide a helpful error message.

PR #2365, opened 2021-07-27
Restrict permissions of ApiUser

mwaldmueller:


Thomas-Gelf:
Rejecting this, as it gives a wrong sense of security. A user equipped with console, config and ticket permissions can give himself whatever permissions he wants at runtime. Also, having API users configured differently on different user systems might cause unexpected issues with upcoming features.

PR #2373, opened 2021-08-12
Update the paths to the IPL and the reactbundle to PHP Thirdparty

theFeu:


Thomas-Gelf:
Documentation should be correct, those requirements are for Icinga Web < 2.9.

theFeu:
From what I have gathered in the forum (I have neither tried it nor looked too much into it tbh) there are some issues with installing the director with Icingaweb > 2.9 due to this

Thomas-Gelf:
This explicitly states `< 2.9.0`, the pull request would break the installation for such versions:

  * If you are using Icinga Web &lt; 2.9.0, the following modules are also required
    * [ipl](https://github.com/Icinga/icingaweb2-module-ipl) (>=0.3.0)
    * [reactbundle](https://github.com/Icinga/icingaweb2-module-reactbundle) (>=0.7.0)

So this section isn't related to 2.9+ at all. Apart from this, installing and enabling those modules doesn't hurt, even with 2.9+. They're just not being used when there are there newer counterparts in place.

theFeu:
I see, fair enough.

PR #2378, opened 2021-08-17
Add `setmembers` summyary line to `TemplateUsageTable` 

raviks789:
Services under service set is shown under `setmembers` of `TemplateUsageTable`. And objects shows only services which do not belong to service set. This must solve the inconsistent number of objects problem.

ref/IP/33703

### DIFF
--- a/library/Director/Db/IcingaObjectFilterHelper.php
+++ b/library/Director/Db/IcingaObjectFilterHelper.php
@@ -5,6 +5,7 @@ namespace Icinga\Module\Director\Db;
 use Icinga\Module\Director\Objects\IcingaObject;
 use Icinga\Module\Director\Resolver\TemplateTree;
 use InvalidArgumentException;
+use Ramsey\Uuid\UuidInterface;
 use RuntimeException;
 use Zend_Db_Select as ZfSelect;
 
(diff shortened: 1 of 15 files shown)


Thomas-Gelf (changes requested):
Hi @raviks789,

showing Services belonging to Service Sets in the "Single Services" list is not a fix for the problem you encountered. The problem is, that the "Usage summary" table doesn't provide a row for Service Templates used by Services in a Set.

[ServiceTemplateUsageTable](https://github.com/Icinga/icingaweb2-module-director/blob/v1.8.1/library/Director/Web/Table/ServiceTemplateUsageTable.php) already shows a related (commented) section. However, this has been never been implemented. The following pieces are missing:

* a route/table showing you such services in a filtered way, together with der Service Set
* an extension to `TemplateUsageTable`, allowing inheriting classes to tweak the underlying query

Best,
Thomas

Thomas-Gelf:
NB: forget what I wrote about `TemplateUsageTable`, it already allows to tweak the query. To fix the numbers for single Services, use:

```php
'objects' => $this->getSummaryLine('object', 'o.service_set_id IS NULL'),
```

To add Service Set members, use:

```php
'setmembers' => $this->getSummaryLine('object', 'o.service_set_id IS NOT NULL'),
```

Then it should look as follows:

director_template_usage_set_members

To then complete the task, a quick attempt could be adding the possibility to filter the Service Set list by "used service template".

Thomas-Gelf:
Cannot be merged yet:

* merge conflict is easy to resolve
* new table predates branches, it does not support them - which is tricky, but a hard requirement right now
* 'objects' summary line needs `o.host_id IS NOT NULL`
* table needs to ignore services bound to hosts (`host_id IS NOT NULL`)
* summary count still does not match table count, probably because on of them counts the services used in our sets, and the other one shows services of sets assigned to single hosts

Thomas-Gelf (approved):

Thomas-Gelf:
Merged, thank you! Disclaimer: I didn't test it at all ;-)

PR #2379, opened 2021-08-20
fix wording of error msg

rndmh3ro:


rndmh3ro:
Please don't make me sign this CLA for this simple fix.

Thomas-Gelf:
> Please don't make me sign this CLA for this simple fix.

Officially, I should. But for such a simple fix I would take the responsibility for breaking the rules ;-)

Regarding the fix: would you suggest to avoid contractions at all, or is there something wrong with `doesn't` in this specific context?

rndmh3ro:
> Officially, I should. But for such a simple fix I would take the responsibility for breaking the rules ;-)

Thanks! :) I'd be fine for me if we'd close this PR and you changed it yourself, too!

> Regarding the fix: would you suggest to avoid contractions at all, or is there something wrong with doesn't in this specific context?

I have nothing against contractions in general. For this case I did it because the output for me on the CLI was:
```
Error message: The related icinga_user "opsgenie" doesn''t exists: Failed to load icinga_user "opsgenie"
```

Notice the two `'`. Since I don't know why there are two of them I just decided to remove the contraction.

Thomas-Gelf:
Thanks for clarification! I would love to figure out why we're messing up the error message. Can you show me, how to trigger this? I tried `icingacli director notification create sample --users missing` - the error message is correct. It should better read "exist" and not "exists", but that's not related to your problem with that doubled quote sign :rofl:

rndmh3ro:
> Can you show me, how to trigger this? 

I use our Ansible [icinga-director-collection](https://github.com/T-Systems-MMS/ansible-collection-icinga-director) for this. And thanks to your question I checked: The double apostrophes come from Ansible. The API returns only one apostrophe.

```
> python3 AnsiballZ_icinga_notification.py execute

{"failed": true, "msg": "bad return code while modifying: 500. Error message: The related icinga_user \"opsgenie\" doesn't exists: Failed to load icinga_user \"opsgenie\"", 
```

So no problem on your side!

PR #2385, opened 2021-08-31
Monitoring: Don't fail if module monitoring is not enabled

nilmerg:
Icinga DB Web loads `monitoring`'s action hooks. If there's no backend configured or the module isn't enabled, Director fails if asked for action urls and runs into a fatal error.

* I'd liked to just catch the error, but that's not possible anymore since PHP 8 (`@` doesn't silence fatal errors anymore)
* I considered to provide a fake backend for a moment, but then noticed class `Monitoring` doesn't even ask for one if the module is disabled
* I don't want to patch Director at runtime

Hence, to make the Director work with Icinga DB Web rc2+ this is required.

### DIFF
--- a/library/Director/ProvidedHook/Monitoring/HostActions.php
+++ b/library/Director/ProvidedHook/Monitoring/HostActions.php
@@ -46,7 +46,7 @@ class HostActions extends HostActionsHook
         $auth = Auth::getInstance();
         if (Util::hasPermission('director/monitoring/hosts')) {
             $monitoring = new Monitoring();
-            if ($monitoring->authCanEditHost($auth, $hostname)) {
+            if ($monitoring->isAvailable() && $monitoring->authCanEditHost($auth, $hostname)) {
                 $allowEdit = IcingaHost::exists($hostname, $db);
             }
         }
--- a/library/Director/ProvidedHook/Monitoring/ServiceActions.php
+++ b/library/Director/ProvidedHook/Monitoring/ServiceActions.php
@@ -56,7 +56,10 @@ class ServiceActions extends ServiceActionsHook
             $title = mt('director', 'Modify');
         } elseif (Util::hasPermission('director/monitoring/services')) {
             $monitoring = new Monitoring();
-            if ($monitoring->authCanEditService(Auth::getInstance(), $hostname, $serviceName)) {
+            if (
+                $monitoring->isAvailable()
+                && $monitoring->authCanEditService(Auth::getInstance(), $hostname, $serviceName)
+            ) {
                 $title = mt('director', 'Modify');
             }
         } elseif (Util::hasPermission('director/monitoring/services-ro')) {


Thomas-Gelf:
Hi @nilmerg,

thanks for reporting this! Do you have a related stack trace for me? Those methods shouldn't be called in such scenario, I would prefer to fix their usage. We might then fail with a `RuntimeException` if called where not available. Returning false feels wrong and might lead to hidden bugs.

nilmerg:
The relevant part is:

```
Uncaught Error: Call to a member function select() on null in /usr/share/icingaweb2-modules/director/library/Director/Monitoring.php:34
Stack trace:
#0 /usr/share/icingaweb2-modules/director/library/Director/Monitoring.php(57): Icinga\Module\Director\Monitoring->hasHost('dummy-0')
#1 /usr/share/icingaweb2-modules/director/library/Director/ProvidedHook/Monitoring/HostActions.php(49): Icinga\Module\Director\Monitoring->authCanEditHost(Object(Icinga\Authentication\Auth), 'dummy-0')
#2 /usr/share/icingaweb2-modules/director/library/Director/ProvidedHook/Monitoring/HostActions.php(21): Icinga\Module\Director\ProvidedHook\Monitoring\HostActions->getThem(Object(Icinga\Module\Icingadb\Compat\CompatHost))
#3 /icingaweb2/modules/monitoring/library/Monitoring/Hook/HostActionsHook.php(50): Icinga\Module\Director\ProvidedHook\Monitoring\HostActions->getActionsForHost(Object(Icinga\Module\Icingadb\Compat\CompatHost))
#4 /icingaweb2/modules/monitoring/library/Monitoring/Hook/ObjectActionsHook.php(21): Icinga\Module\Monitoring\Hook\HostActionsHook->getActionsForObject(Object(Icinga\Module\Icingadb\Compat\CompatHost))
```

Thomas-Gelf:
Thank you, this helps! This isn't triggered by the Director, but from the Hook that is being called from the monitoring module. This shouldn't be called at all, when the monitoring module isn't active. Blind guess: you're calling monitoring module Hooks from the IcingaDB web module. You can prevent this error by calling isAvailable() immediately after `new Monitoring` in Host- and ServiceActions.php.

In addition to this, I'd strongly suggest to reconsider running monitoring module Hooks by it's successor. Reasons:

* as this example shows, modules providing Hooks might make the assumption, that the monitoring module is there. They'll fail when calling related code.
* the new module will provide better Hooks over time, as it has more potential for cool integrations - thanks to (not only) ipl\Html. Module authors will try to provide Hooks for both the IcingaDB and the IDO monitoring module, with fancier ones for the new world. They risk to show up twice in the new world.

I understand, that enabling old Hooks in a compatible way eases migration and speeds things up. Still, I'm not sure whether this is the right path to follow. You might want to get a second opinion on this.

Apart from that decision, please adjust the Pull request - I'll merge it immediately. Placing it into the Hooks would fix my above concerns.

Thanks,
Thomas

nilmerg:
> as this example shows, modules providing Hooks might make the assumption, that the monitoring module is there. They'll fail when calling related code.

I'm personally also not a fan of this, DB Web has its own hooks now. But it's been requested still.

> They risk to show up twice in the new world.

This is addressed and won't happen.

Thomas-Gelf:
Thanks!

PR #2387, opened 2021-09-08
Allow macros to be a dict in service-apply-rules

### LINKED ISSUE #2083: Using macro sub-properties not working in Service Apply Rules

### LINKED ISSUE #2432: Allow for more then just $config$ in Service Apply For Rule properties
## Expected Behavior
I have an data field looking like something like this:
```
 {
    "name" => "Name".
    "parameter1" => 123,
    "parameter2" => 456
 }
```
I use it in an Service Apply Rule with Apply For.

I set the following:
Service Name: `Check $config.name$`
Custom Properties: Property 1 => `$config.parameter1$` and Property 2 => `$config.parameter2$` 

I expect the Service to be named `Check Name` and the properties to be set to `123` and `456` respectively.

## Current Behavior
The Service gets named correctly.
The two properties will be populated with the String `"$config.parameter1$"` and `"$config.parameter2$"`.

The check will report as UNKNOWN with this error:
`Error: Non-optional macro 'config.parameter1' used in argument 'parameter1' is missing.`

It seems the macro is expanded at an later point, which is not possible anymore since context is missing.

## Possible Solution
Allow for specifying parameters in Custom Properties. At the moment only `$config$` is allowed in https://github.com/Icinga/icingaweb2-module-director/blob/master/library/Director/CustomVariable/CustomVariableString.php#L49 but this will fill in my entire object.

## Your Environment
* Director version (System - About): 1.8.0
* Webserver, PHP versions: PHP 8.0.13

Yoda-BZH:
Same as #1453,  #2085 or #2446

In Service-Apply-Rules macros like `$config.property$` should be evaluated, currently the only macro that gets evaluated is `$config$`

also fixes #2083, fixes #2432.

This method doesn't need `toConfigString` modification.

This PR contains tests against this modification.


Yoda-BZH:
verified.

bobapple:
@cla-bot check

Yoda-BZH:
thank you.

Is there a way to run the CI against my PR ?

Yoda-BZH:
any news ?

jcourreges:
Hi there.

we're still depending on this diff at $DAYJOB. How can we help getting this in?

PR #2401, opened 2021-09-30
Convert $imports keys to string explicitly

raviks789:
The keys of $imports in IcingaObjectImports must be converted to string explicitly or else it establishes incorrect parent-child relation
in icinga_host_inheritance table in case the key is integer value and there is an host object with id = key(integer) icinga_host table.
This results in undefined offset while triggering sync rule.

[ref/IP/36089](https://rt.icinga.com/Ticket/Display.html?id=36089&results=9fd14bd96eccc7871e1fbc5f10bf4b96)

### DIFF
--- a/library/Director/Objects/IcingaObjectImports.php
+++ b/library/Director/Objects/IcingaObjectImports.php
@@ -232,6 +232,7 @@ class IcingaObjectImports implements Iterator, Countable, IcingaConfigRenderer
     {
         $list = [];
         foreach ($this->listImportNames() as $name) {
+            $name = (string) $name;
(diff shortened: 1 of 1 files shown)


Thomas-Gelf:
Merged, as it doesn't hurt. But please note that I STRONGLY advise AGAINST using numeric Template names. Whoever came up with such a terribly idea should be punished ;-) Please add some String-Prefix to those names. As of how array-indexing-magic works, we can (and will not) guarantee that this will work at all.

We should better introduce a warning for such objects in the next major release, and start throwing errors at some later point in time.

Anyways, thank you!

PR #2402, opened 2021-09-30
Pass $resolved as False for arguments to toPlainObject()

raviks789:
IcingaArguments passes the $resolved through to the individual argument in the toPlainObject. There lies the error,
simply pass false - because the argument itself cannot have any parents and therefore cannot be resolved

[ref/IP/36053](https://rt.icinga.com/Ticket/Display.html?id=36053)


Thomas-Gelf:
Thanks!

PR #2403, opened 2021-10-06
Add feature to specify a custom endpoint name for a host

lazyfrosch:
- Render Endpoint and Zone with a different name per host
- Add custom variable `_director_custom_endpoint_name` to a host with that name
- Update `command_endpoint` behavior in services to use custom var or hostname
- Includes a feature flag that needs to be enabled

## Screenshots

host-form

host-preview

service-after

settings

## Tasks

- [x] Update schema, lock migration ID

cc @mkayontour 
ref/NC/729576

### DIFF
--- a/library/Director/Objects/IcingaHost.php
+++ b/library/Director/Objects/IcingaHost.php
@@ -252,14 +252,11 @@ class IcingaHost extends IcingaObject implements ExportInterface
         $config->configFile($pre . 'agent_zones')->addObject($zone);
     }
 
-    protected function renderCustom_endpoint_name()
+    protected function renderCustom_endpoint_name($value)
     {
         // When feature flag feature_custom_endpoint is enabled, render custom var
         if ($this->connection->settings()->get('feature_custom_endpoint') === 'y') {
-            return c::renderKeyValue(
-                'vars._director_custom_endpoint_name',
-                c::renderPhpValue($this->get('custom_endpoint_name'))
-            );
+            return c::renderKeyValue('vars._director_custom_endpoint_name', c::renderString($value));
         }
 
         return '';


lazyfrosch:
@Thomas-Gelf Any opinion on this?

Thomas-Gelf:
Hi @lazyfrosch,

thank you for this contribution! Thoughts:

* no setting required, this should be available all the times
* show the related field only for admins, with an according note stating that changing this property it is VERY easy to shoot yourself in the foot. You cannot choose your endpoint at will, as the related Host object MUST be present in the Zone you're sending your service to.
* Restricted configuration possibilities are easy for single service objects, but would become tricky plausibility checks for templates - so that's not an easy task. Which means, the user must be fully aware of what he's doing
* I do not understand the necessity for the hack with that custom variable. Why not linking an Endpoint at Service (Template) level? I'm sure you have good reasons for this. I haven't been able to find related information in the linked ticket. An example use-case would be helpful, as w/o understanding the reasoning behind the feature, this feels quite hacky. But I'm always open for helpful hacks!
* Defined this way, how would you run a "Syslog Check" for host A on your Syslog Server Endpoint, and a "DNS Plausibility Check" for host A on a dedicated Lookup Server Endpoint?
* endpoint_host_id would feel more correct to me. Related lookups should take place at rendering time
* this feature might have to wait: we recently added UUIDs to most object tables, UUID-based relation tables and -properties will follow. As UUIDs will give us more possibilities, I'm currently trying to fade out shady text-based relations. This pull-request might as of this writing add a new (undesired) text-based relation, while it could instead benefit a lot from UUID-based relations once available
* please remove unrelated formatting changes from this pull request. Where necessary I'd welcome a dedicated pull request

Thanks,
Thomas

lazyfrosch:
Cleaned up unrelated changes.

moreamazingnick:
If I understand that correctly this is more a renaming feature.

But what if we introduce a host_setting (run_on_other_agent , boolean)
and if that set to yes we say that a agent/zone should not be rendered if run_on_other_agent === true
```
       if($this->get('custom_endpoint_name') !== null && $this->get('run_on_other_agent') === true ){
            return;
       }
```

I added the skipping part as a PoC, and it works, but I dont want to mess with your database

I would be happy if this finds a way into icigna director the right way.

https://github.com/moreamazingnick/icingaweb2-module-director/tree/poc_feature/real_custom_endpoint

PR #2405, opened 2021-10-07
Fix and enhance localization markings

nilmerg:
Requires ipl-web v0.4.0. There may be more issues, didn't check further. The ipl-web requirement will break with non-v2.9 web installations. So merge it or not :man_shrugging:


Thomas-Gelf:
Thanks for pointing this out, merged a simplified version w/o new dependencies

PR #2420, opened 2021-10-28
Restore services of service set on restoring service set.

### LINKED ISSUE #1065: Service Sets aren't fully restored if one is accidentally deleted.
If I accidentally klick on the delete Button of a service set and try to "restore former object" only an empty service set is restored. 

"Restore former object" should restore the complete director configuration.

Create a service set "test_service_set" and add one or more services to it
Deploy Config
delete service set "test_service_set"
Go to Acitivity log and click on "Restore former object"

The service set is recreated empty.

## Your Environment
* Director version (System - About): 1.3.1
* Icinga Web 2 version and modules (System - About): 2.4.1
* Icinga 2 version (`icinga2 --version`): v2.6.3
* Operating System and version: Red Hat Enterprise Linux Server 7.3
* Webserver, PHP versions: httpd-2.4.6-45.el7 php-5.4.16-42

raviks789:
All the services which were under the service set when it was deleted must be restored when it is restored from activity log.

dependent on PR: #2418 

fix #1065

### DIFF
--- a/library/Director/Web/Widget/ActivityLogInfo.php
+++ b/library/Director/Web/Widget/ActivityLogInfo.php
@@ -3,7 +3,10 @@
 namespace Icinga\Module\Director\Web\Widget;
 
 use gipfl\Json\JsonString;
+use Icinga\Module\Director\Data\FieldReferenceLoader;
+use Icinga\Module\Director\DirectorObject\Automation\BasketSnapshotFieldResolver;
 use Icinga\Module\Director\Objects\DirectorActivityLog;
+use Icinga\Module\Director\Web\Form\IcingaObjectFieldLoader;
 use ipl\Html\HtmlDocument;
 use ipl\Html\HtmlElement;
 use Icinga\Date\DateFormatter;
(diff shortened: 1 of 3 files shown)


lippserd:
ref/IP/35863

Thomas-Gelf:
Hi @raviks789,

thank you for this PR! Main problems:

* it doesn't work with PostgreSQL (can be fixed)
* it's a wild hack (per design, that's not your fault - it's a tricky task)

You correctly figured out, that  related Service activities must have their Activity (Removal) Log somehow being triggered. I'd have had preferred a direct call to `DirectorActivityLog::logRemoval()` for each service - depending on a specific `onDelete`-Implementation is fragile.

But please let me propose a different implementation. The main issue here is how deletion takes place. A user issues a single operation ("Delete a Service Set"), but behind the scenes, multiple operations take place. For a better end user experience, this should be "one step" throughout the UI. Read: there should be only one entry in the Activity Log.

This sounds tricky, but the good news is: most of the related work has already been done. `IcingaServiceSet` implements the `ExportInterface`, and therefore has import/export-Methods. They're being used by our Configuration Baskets, and have one big advantage: no fiddling with IDs on restore. It is up to you to figure out, how this then fits into our current Show-Config-Diff-Implementation ;-)

Could you please give that approach a try? Only if you find the time of course, otherwise please let me know. I'm allowed to dedicate some time to similar tasks this week.

Thanks a lot,
Thomas

Thomas-Gelf (dismissed):
.

tbauriedel:
ref/NC/732413

nilmerg (commented):
The export/import logic wasn't removed, only moved. (56f81b91bb0cd4f8be6e9d2aa11509659c5620a5) Isn't it still applicable here?

library/Director/Web/Widget/ActivityLogInfo.php
```diff
@@ -624,10 +643,31 @@ protected function createObject($type, $props)
...
             $this->db
-        )->setProperties((array) $props);
+        );
+
+        if ($type === 'icinga_service_set' && isset($props->services)) {
+            $services = [];
+            foreach ($props->services as $service) {
+                $serviceProps = ['object_name' => $service->object_name];
+                $serviceProps['object_type'] = $service->object_type;
+                $serviceProps['imports'] = $service->imports;
```
nilmerg, line 657: Please make sure all configurable properties are restored.

nilmerg (changes requested):

nilmerg (approved):

PR #2423, opened 2021-11-04
schema: Quote keyword `groups`

nilmerg:
It's a keyword on MySQL 8 now. (https://dev.mysql.com/doc/refman/8.0/en/keywords.html)


Thomas-Gelf:
Thank you!

PR #2436, opened 2021-11-24
Add cube links for objects from Icingadb and register the hook

raviks789:
fixes Icinga/icingaweb2-module-cube#64


Thomas-Gelf:
I didn't test it, but the code looks good to me - thank you!

PR #2438, opened 2021-11-24
Use hosts deployed from director as url parameters in CubeLinks

### LINKED ISSUE #2437: CubeLink doesnot work in case the hosts are not deployed from director.
## Expected Behavior
Only the hosts deployed from director must be used as parameters for CubLink

## Screen Recording

https://user-images.githubusercontent.com/33730024/143267781-25385e9b-0836-4ec5-90e3-f176f3709e62.mov

## Possible Solution
Remove the hosts from the url parameter list that were not deployed from director.

raviks789:
In case there were none of the hosts showed in a dimension for the cube module, then do not show the CubeLinks for that dimension.

fixes #2437

### DIFF
--- a/library/Director/ProvidedHook/CubeLinks.php
+++ b/library/Director/ProvidedHook/CubeLinks.php
@@ -2,10 +2,13 @@
 
 namespace Icinga\Module\Director\ProvidedHook;
 
+use Icinga\Application\Config;
 use Icinga\Data\Filter\Filter;
 use Icinga\Module\Cube\Cube;
 use Icinga\Module\Cube\Hook\ActionsHook;
 use Icinga\Module\Cube\Ido\IdoHostStatusCube;
+use Icinga\Module\Director\Db;
+use Icinga\Module\Director\Objects\IcingaObject;
 use Icinga\Web\View;
 
 class CubeLinks extends ActionsHook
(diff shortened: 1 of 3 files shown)


library/Director/ProvidedHook/CubeLinks.php
```diff
@@ -19,6 +22,8 @@ public function prepareActionLinks(Cube $cube, View $view)
             return;
         }
 
+        $directorHosts = array_keys(IcingaObject::loadAllByType('Host', $this->directorDb()));
```
Thomas-Gelf, line 25: This is not a good idea, loading all Hosts is not an option. While the user might select just a bunch of them, the DB can easily contain 100k+.

Thomas-Gelf (changes requested):

library/Director/ProvidedHook/CubeLinks.php
nilmerg: That's already the case here: https://github.com/Icinga/icingaweb2-module-director/pull/2436/files#diff-37cfa9169f450b93bda4663c3e82742f0d45336115ae2394ab9c831cab184be9R29, Just FYI

Thomas-Gelf: @nilmerg: thanks for pointing this out, I overlooked it! @raviks789: please either use the deployment location in IcingaDB as a source of truth or ask the Director about the selected hosts only. If unsure how to address this, please let me know

raviks789:
This was not worked upon for a long time and the fix seems to be incomplete, so closing this for now.

PR #2445, opened 2021-12-02
Enrich custom variable lists in monitoring and icingadb details

### LINKED ISSUE #2239: Show Director Labels in Monitoring Module Custom Vars section
## Expected Behavior

Custom vars added to a host via Icinga Director employ the key defined in Director but not the label which is what Director displays in all its forms. For consistency and because labels tend to be more human-friendly I'd like to have Director override this behaviour.
It will be enough to have it as an optional setting.

## Current Behavior

If we create a data list in Director such as.
Example Data List "Location":

- Key / Label

- bcn / Barcelona
- lon / London
- par/ Paris

We can now assign a "Location" value to a host selecting from a dropdown (Barcelona, London, Paris) and that label will be what all that you see in Director templates/host forms/etc....
However in the monitoring module when you click on the host you will see that the custom var "location" has the value "bcn/lon/par" which is not ideal particulary because there is no easy way to rename existing keys without deleting/readding.

## Possible Solution

In Icinga Insights chat there was a suggestion that there is a hook on the monitoring module that will allow Director to override the usage of keys for labels.

## Your Environment

* Director version (System - About): 1.7
* Icinga Web 2 version and modules (System - About): 2.8.1
* Icinga 2 version (`icinga2 --version`):2.12
* Operating System and version: Centos 8
* Webserver, PHP versions: Apache 2.4 + PHP 7.2

nilmerg:
resolves #2239

requires https://github.com/Icinga/icingaweb2/pull/4614 or https://github.com/Icinga/icingadb-web/pull/467

### DIFF
--- a/library/Director/ProvidedHook/Icingadb/CustomVarRenderer.php
+++ b/library/Director/ProvidedHook/Icingadb/CustomVarRenderer.php
@@ -12,6 +12,12 @@ use Icinga\Module\Director\Web\Form\IcingaObjectFieldLoader;
 use Icinga\Module\Icingadb\Hook\CustomVarRendererHook;
 use Icinga\Module\Icingadb\Model\Host;
 use Icinga\Module\Icingadb\Model\Service;
+use ipl\Html\Attributes;
+use ipl\Html\Html;
+use ipl\Html\HtmlDocument;
+use ipl\Html\HtmlElement;
+use ipl\Html\Text;
+use ipl\Html\ValidHtml;
 use ipl\Orm\Model;
 
 class CustomVarRenderer extends CustomVarRendererHook
@@ -24,6 +30,14 @@ class CustomVarRenderer extends CustomVarRendererHook
 
     /** @var array Related dictionary field names */
     protected $dictionaryNames = [];
+
+    protected $dictionaryLevel = 0;
+
+    /** @var HtmlElement Table for dictionary fields */
+    private $dictionaryTable;
+
+    /** @var HtmlElement Table body for dictionary fields */
+    private $dictionaryBody;
 
     /**
      * Get a database connection to the director database
@@ -126,7 +140,11 @@ class CustomVarRenderer extends CustomVarRendererHook
     public function renderCustomVarKey(string $key)
     {
         if (isset($this->fieldConfig[$key]['label'])) {
-            return $this->fieldConfig[$key]['label'];
+            return new HtmlElement(
+                'span',
+                Attributes::create(['title' => $this->fieldConfig[$key]['label'] . " [$key]"]),
+                Text::create($this->fieldConfig[$key]['label'])
+            );
         }
 
         return null;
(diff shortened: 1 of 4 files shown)


Thomas-Gelf (changes requested):
I really like the motivation behind this feature request and the possibilities this hook provides.

Problems spotted in this pull request: field resolution cannot be solved by joining field names with variable names. It's perfectly legal to configure the very same field multiple times, with different labels, different settings/rules and in different categories. Fields might also be provided via Hook.

This means that walking the template tree in the right order, dealing with fields with filters and fields overriding fields at template level are part of this task.

To avoid duplicate logic, you could leverage IcingaObjectFieldLoader, as it is doing exactly this. I'd suggest to run & populate it, then get the form elements and pick their labels and values. If you need to restructure that class please let me know, I'm sure I can help you saving some time with that task.

nilmerg:
I've made use of `IcingaObjectFieldLoader` now, thanks for this hint. Though, `IcingaObjectFieldLoader::getFields()` already seems to work fine with the template tree and hooks, no need to access any form elements. :thinking:

nilmerg:
Please don't merge yet, or give me a hint when you plan to do so. I'm currently playing with an idea to extend this to data exports. (So that your hidden option for strings also works in json and csv exports)

nilmerg:
> Please don't merge yet, or give me a hint when you plan to do so. I'm currently playing with an idea to extend this to data exports. (So that your hidden option for strings also works in json and csv exports)

Nope. Changed my mind. Nothing to do here. You can merge, if you are fine with the changes.

Thomas-Gelf:
Hi @nilmerg,

I discussed this thoroughly with @lippserd, there are some problems with this Hook. The variable rendering part is fine, the way variables are iterated is the problem. We already agreed on a slightly different implementation, with only some smaller details left for clarification. We also had a customer interested in pushing the finalization of this, it was part of a deal with a vSphereDB-related feature.

Unfortunately, this hasn't been accepted from our side for unknown reasons, so I have to skip this right now. We'll find a solution, don't worry - but not today.

In the meantime, thank you for your effort!

Cheers,
Thomas

library/Director/ProvidedHook/Icingadb/CustomVarRenderer.php
```diff
@@ -130,6 +135,13 @@ public function renderCustomVarValue(string $key, $value)
 
             if (isset($this->datalistMaps[$key][$value])) {
                 return $this->datalistMaps[$key][$value];
+<<<<<<< Updated upstream
+            } elseif (in_array($key, $this->dictionaryNames)) {
+                return $this->renderDictionaryVal($value);
+=======
+            } elseif ($value !== null && in_array($key, $this->dictionaryNames)) {
+                return $this->renderDictionaryVal($key, (array) $value);
+>>>>>>> Stashed changes
```
nilmerg: please amend this, syntax errors shouldn't remain anywhere in commits

nilmerg: Both commits before the phpstan baseline update seem to be duplicates of those right after the baseline update. Please merge them.

PR #2446, opened 2021-12-06
Allow macros with parameters

### LINKED ISSUE #2432: Allow for more then just $config$ in Service Apply For Rule properties

v0tti:
This fixes #2432 by adding a `matchRegex` parameter. It should not break any existing use cases and just allows for more flexibility when using "Apply For" rules.


v0tti:
I added the commit email address to my GitHub account, should be good now.

nilmerg:
@cla-bot check

v0tti:
I have signed the CLA now.
@cla-bot check

bobapple:
@cla-bot check

PR #2460, opened 2022-01-14
Avoid passing non-string args to `ctype_*()` functions

nilmerg:
https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.ctype


Thomas-Gelf:
Merged, thank you! Just curious, have you been able to trigger them - or did you "just" grep for `ctype_*`?

While it's not clear from the code, the first one SHOULD always get a string parameter, as it's values come from ZF form fields (empty strings SHOULD be transformed to null later on), and the second one (valueToTimestamp) SHOULD have been left over from a former use - and never be triggered in the Director.

PR #2463, opened 2022-01-21
Added possibility of using functions in fields of type string

mlg1:
Strings starting with "@@" will be rendered without quotes. This will allow us to use the Icinga2 functions in the director.
https://icinga.com/docs/icinga-2/latest/doc/18-library-reference/#object-accessor-functions


mlg1:
I signed the agreement after I did a pull request. Would you please check?

lippserd:
@cla-bot check

mlg1:
Thank you! I added another commit that fixes a problem with the use of special characters.

PR #2464, opened 2022-01-24
Delete newly added Services in Service Set when restoring it from snapshot

raviks789:
The Services which were added into the Service Set after the snapshot was created must be deleted when the Service Set is being restored from the snapshot.

[ref/IP/37937](https://rt.icinga.com/Ticket/Display.html?id=37937)


Thomas-Gelf:
Merged, thank you!

PR #2465, opened 2022-01-24
Compatibility with ipl v0.8.0

nilmerg:
With the next non-patch release of the IPL, it will require PHP 7.2 at a minimum and adds support for up to PHP 8.1. This means that some interfaces will change. For example `ipl\Html\HtmlDocument::count()` will change to `ipl\Html\HtmlDocument::count(): int`. This is addressed here for the `Icinga\Module\Director\Dashboard\Dashboard` abstract class.

This also means that the Director has to raise it's minimum required version of Icinga Web 2 up to v2.9, as otherwise it's not possible to install the IPL in v0.8.0.

I've also thought about just not overriding `ipl\Html\HtmlDocument::count(): int` and instead patch `Icinga\Module\Director\Dashboard\Dashboard::isAvailable()` as that wouldn't require any dependency upgrades, as far as I know, since this is the only case I've identified.

But since there are also some cases in `gipfl\`, I figured it isn't worthwhile since I don't think a similar work-around is possible there.

### DIFF
--- a/doc/02-Installation.md
+++ b/doc/02-Installation.md
@@ -16,17 +16,15 @@ Requirements
   * Some features require newer Icinga 2 releases
     * Flapping requires 2.8 for the thresholds to work - and at least 2.7 on all
       nodes
-* Icinga Web 2 (&gt;= 2.6.0). All versions since 2.2 should also work fine, but
-  might show smaller UI bugs and are not actively tested
+* Icinga Web 2 (&gt;= 2.9.0)
+* [Icinga PHP Library (ipl)](https://github.com/Icinga/icinga-php-library) (>= 0.8)
+* [Icinga PHP Thirdparty](https://github.com/Icinga/icinga-php-thirdparty) (>= 0.10)
 * The following Icinga modules must be installed and enabled:
   * [incubator](https://github.com/Icinga/icingaweb2-module-incubator) (>=0.5.0)
-  * If you are using Icinga Web &lt; 2.9.0, the following modules are also required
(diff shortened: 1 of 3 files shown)


lippserd:
Needs also a fix in gipfl:

```
Fatal error: Declaration of gipfl\IcingaWeb2\Table\QueryBasedTable::count() must be compatible with ipl\Html\HtmlDocument::count(): int in /usr/share/icingaweb2-modules/incubator/vendor/gipfl/icingaweb2/src/Table/QueryBasedTable.php on line 64

Thomas-Gelf:
@lippserd: IMHO we should not breaking base ipl classes, which have been designed for being inherited from. This breaks a bunch of modules, I'd instead suggest supporting PHP 8.1 in a compatible way. As explained back in [November](https://github.com/Icinga/icingaweb2/issues/4609#issuecomment-977622769), `ReturnTypeWillChange` is the way to go if you want to support a wide range of versions. As of this writing, for `Countable` this means 5.1-8.4.

I'm all in for a renewed version of `ipl\Html` once the time is ready, but strictly against breaking things again and again. I see no gain in shipping alternative/updated interfaces unless we require at least PHP 7.4.

Said all this, please revert the change in `ipl\Html` and add `#[\ReturnTypeWillChange]`, this has only advantages. And don't worry: this does not imply reverting related changes in other modules, where related modifications (like adding `: int`) already took place. They'll continue to work fine.

lippserd:
Will be handled with https://github.com/Icinga/ipl-html/pull/62.

Thomas-Gelf:
Great, thank you!

PR #2467, opened 2022-01-28
Fix sync rule restore from snapshot on name change

raviks789:
Icinga director throws error when we try to restore sync rule from snapshot of the configuration basket after its name has been changed.

[ref/IP/37949](https://rt.icinga.com/Ticket/Display.html?id=37949#txn-638589)

### DIFF
--- a/library/Director/Objects/SyncRule.php
+++ b/library/Director/Objects/SyncRule.php
@@ -315,10 +315,6 @@ class SyncRule extends DbObject implements ExportInterface
         $object->newSyncProperties = $properties['properties'];
         unset($properties['properties']);
         $object->setProperties($properties);
-        if ($id !== null && (int) $id !== (int) $object->get('id')) {
-            $object->originalId = $object->get('id');
-            $object->reallySet('id', $id);
-        }
 
         return $object;
     }


Thomas-Gelf:
Hi @raviks789,

thanks a lot! The second part of this pull request scares me a little bit, however I didn't investigate whether it has or hasn't any side effects. Anyway, the culprit is elsewhere: I do not like what `SyncRule::import()` does with IDs. Guess the original idea was to "preserve IDs, at least where possible". Didn't work out as intended, that's why other import methods are not doing so. I'll check whether that special treatment can be removed an let you know, that would fix the problem at it's root.

Cheers,
Thomas

wp-perc:
Hi @Thomas-Gelf, I am the source of this headache of yours :)

The issue behind this pull request is a good blocker for us.
Do you think will be merged anytime soon?

Thomas-Gelf:
Hi @wp-perc,

I'm in contact with @raviks789. The pull request will not be merged as it is, he'll provide a modified version. On name change, a new Sync Rule will be created. We're not going to deal with IDs, as this is too dangerous once multiple systems are involved.

In future we'll bring UUIDs to non-Icinga-objects (and to our baskets) too, and then we'll allow baskets to also rename objects.

Cheers,
Thomas

Thomas-Gelf (approved):
LGTM, thank you!

PR #2468, opened 2022-02-01
Prevent deletion of data lists that are in use.

raviks789:
Deletion of data lists which are used in icinga objects through custom variables is prevented. As the objects that use these data lists could not be restored through snapshots otherwise.

[ref/IP/37953](https://rt.icinga.com/Ticket/Display.html?id=37953)


Thomas-Gelf:
LGTM, thank you!

PR #2476, opened 2022-02-18
Keep a single CubeLinks hook.

raviks789:
An additional IcingaDbCubeLinks hook is unnecessary. Remove this and make changes to
CubeLinks so that it works in case of icingadb as well.


nilmerg:
As mentioned this morning, we shouldn't change the legacy hook, in order to keep maximum compatibility.

PR #2484, opened 2022-03-02
Cleanup IcingaDbCubeLinks with getObjectsFilter provided by IcingaDbCube.

raviks789:
IcingaDbCubeLinks needs to be cleaned up. The method getObjectsFilter() return the needed filter query for the action link.
Also, retrieving hosts from director to have a friendly fail for action link is incorrect. This must be handled by ObjectsController and that would be a different pull request.


nilmerg (commented):
Looks fine so far. Though, it still misses handling for non-existent hosts just like the legacy hook. (But I don't care if that's handled here or separately.)

PR #2491, opened 2022-03-09
Fix UTF8 encoding problem for baskets and remove UUIDs from the basket

### LINKED ISSUE #2488: Basket with simple Service set contains Malformed UTF-8 characters
## Current Behavior
Creation of basket snapshot fails if Service set is included:
basket1

Example Service set:
serviceset1

## Your Environment

* Director version (System - About): 1.9.0
* Icinga 2 version (`icinga2 --version`): 2.13.2
* Operating System and version: EL8
* Webserver, PHP versions: Apache, PHP 7.2.24

RincewindsHat:
- Ugly fix to restore basket export
- Remove UUIDs from basket since they are not transferable between different setups

Fixes #2488

### DIFF
--- a/library/Director/DirectorObject/Automation/BasketSnapshot.php
+++ b/library/Director/DirectorObject/Automation/BasketSnapshot.php
@@ -199,6 +199,7 @@ class BasketSnapshot extends DbObject
             throw new RuntimeException('A basket snapshot cannot be modified');
         }
         $json = $this->getJsonDump();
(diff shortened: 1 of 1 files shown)


RincewindsHat:
Should be fixed by [11f621df30ad65b39f203863530189b15dc5174e](https://github.com/Icinga/icingaweb2-module-director/commit/11f621df30ad65b39f203863530189b15dc5174e)

PR #2502, opened 2022-03-28
fixes #2393 clusterzone check is not a plugincheck

### LINKED ISSUE #2393: IcingaDirector can't create ClusterZoneCheck
https://github.com/Icinga/icingaweb2-module-director/blob/302c9b645d901567df5433657487701627396d0d/library/Director/Objects/IcingaCommand.php#L63
should be 
        'ClusterZoneCheck' => 'cluster-zone-check-command',

thanks for this great software and

best regard
nicolas

moreamazingnick:


lippserd:
@raviks789 please check PR.

moreamazingnick:
Somehow it works without the change since 1.10

lippserd:
> Somehow it works without the change since 1.10

Thanks for the reply. But I think it should still be corrected to `cluster-zone-check-command`.

raviks789 (commented):
LGTM

Issue #2505, opened 2022-03-28
Cleared activity log causes exception


moreamazingnick:
Undefined offset: 0 (ActivityLogTable.php:65)
#0 /usr/share/icingaweb2/modules/director/library/Director/Web/Table/ActivityLogTable.php(65): ...

https://github.com/Icinga/icingaweb2-module-director/blob/9afa3313ab9f9e07f8f1f3b7915bb52f340ab176/library/Director/Web/Table/ActivityLogTable.php#L65

PR #2506, opened 2022-03-28
fix empty activity log causes exception refs #2505

moreamazingnick:


library/Director/Web/Table/ActivityLogTable.php
```diff
@@ -62,6 +62,9 @@ protected function fetchQueryRows()
     {
         $rows = parent::fetchQueryRows();
         // Hint -> DESC, that's why they are inverted
+        if( empty($rows)){
```
Thomas-Gelf, line 65: if (empty($rows)) {

Thomas-Gelf (changes requested):
Thank you! Looks good to me, please fix the syntax as shown (phpcs should not complain) and amend/squash/force-push/whatever :-)

Issue #2456, opened 2022-01-05
director creates wrong url for scheduled downtime in activity log


moreamazingnick:
https://github.com/Icinga/icingaweb2-module-director/blob/087ce05ebe9ec935d14ceb73293a2ce25b8088b5/library/Director/Web/Table/ActivityLogTable.php#L79

wrong url
https://icinga-master/icingaweb2/director/config/activities#!/icingaweb2/director/scheduleddowntime?name=THISDowntime

right url
https://icinga-master/icingaweb2/director/config/activity?id=12234#!/icingaweb2/director/scheduled-downtime?name=THISDowntime

moreamazingnick:
@Thomas-Gelf still present in 1.10
It happens if you klick on the name url => hostwithservices
[nicolas] [modify] scheduled downtime ["hostwithservices"](https://icinga.master/icingaweb2/director/scheduleddowntime?name=hostwithservices)

slalomsk8er:
just stumbled over the same bug

PR #2507, opened 2022-03-28
fixes wrong url in activitylog for scheduledowntime #2456

moreamazingnick:


library/Director/Web/Table/ActivityLogTable.php
```diff
@@ -213,6 +213,13 @@ protected function makeLink($row)
...
+            } elseif ($type === 'scheduled_downtime') {
+                $object = Link::create(
+                    "\"$name\"",
+                    'director/' . str_replace('_', '-', $type),
+                    ['name' => $name],
+                    ['title' => $this->translate('Jump to this object')]
+                );
             } else {
                 $object = Link::create(
                     "\"$name\"",
```
raviks789, line 225: I would instead handle this in the `else` statement like below. Otherwise, the solution looks acceptable to me.

```suggestion
            } else {
                $delimiter = $type === 'scheduled_downtime' ? '-' : '';
                $object = Link::create(
                    "\"$name\"",
                    'director/' . str_replace('_', $delimiter, $type),
                    ['name' => $name],
                    ['title' => $this->translate('Jump to this object')]
                );
            }
```

lippserd: @moreamazingnick I tried to force push, but unfortunately I don't have the rights to do so.

moreamazingnick:
I think you already have access to edit, the static errros are intoduced by commit fe1c7c92f0566e6cca9fc5d20ede16f429eb35d8

raviks789:
> I think you already have access to edit, the static errros are intoduced by commit [fe1c7c9](https://github.com/Icinga/icingaweb2-module-director/commit/fe1c7c92f0566e6cca9fc5d20ede16f429eb35d8)

This happened while rebasing the branch here in github. But I am unable to push commits to this PR.

Issue #608, opened 2016-11-17
[dev.icinga.com #13233] Missing "display_name" field in Service tab


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/13233

**Created by plarivee on 2016-11-17 18:56:02 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-12-04 13:43:08 +00:00 (in Redmine)_

---

there is no "display\_name" field in the service tab when creating / editing services

vita2:
it would be nice to have the display_name argument as a optional field :+1:

druchoo:
I'm trying to create the Cluster Zone check(s) as described [here](https://www.icinga.com/docs/icinga2/latest/doc/06-distributed-monitoring/#health-checks). Is there another way to do this with a dynamically generated name without display name?

```
apply Service "cluster-health" {
  check_command = "cluster-zone"

  display_name = "cluster-health-" + host.name

  /* This follows the convention that the client zone name is the FQDN which is the same as the host object name. */
  vars.cluster_zone = host.name

  assign where host.vars.client_endpoint
}
```

bahrpatrick:
I would really appreciate this feature in the Icinga2 director.
Is there any workaround to set the "display_name" at the moment?

Hitokun:
for anyone wondering, like @bahrpatrick. You can get the id of the service apply (look at your browser's address bar for it when looking at the apply in director, eg: http://172.17.0.2/icingaweb2/director/service?id=88).

With that reference you can POST to the API like (forgive my pseudo-curl) `curl -s user:pass -XPOST -H content-type:application/json http://myicinga.com/icingaweb2/director/service?id=<id> -d '{"display_name": "My Service Display Name"}'`. It is not hard to do, but it is pretty time consuming.

My use case involves having different apply rules (where the custom fields are different) where I want the same display name among them all.

I do agree this would be pretty cool to have in place. I'm not very familiar with Director's codebase, so I don't know how hard it would be to add the field to the Apply Rule view (assuming that's the only thing required). I might spare a couple of hours one of these days (no guarantee's, though) and try to add this. Would be a fun :grimacing:.

rcalpha:
[...]

PR #2509, opened 2022-03-30
Feature/unlock display name #608

moreamazingnick:


raviks789:
Thanks for the contribution, and sorry this sat without review for so long.

Checking the current state of master, the display name work here has partly landed already through #3078

  - The one piece that's still missing is ObjectsTableService::$searchColumns, which doesn't include display_name for services (only object_name for both service and host).

Since most of this PR's intent is already covered I'm closing this PR.

PR #2524, opened 2022-04-19
Prevent inserting blank entries for extensible data list.

raviks789:
Before this change, when the value for the data field of data type "Datalist" (and element behaviour is "Extend the list with new values") is blank, then the director throws mysql error for trying to insert a blank entry into data list entry table. 

[ref/IP/33703](https://rt.icinga.com/Ticket/Display.html?id=38958)


Thomas-Gelf:
Merged, thank you!

PR #2526, opened 2022-04-20
PropertyModifier: Replace Null with String

pdolinic:
Because of recent fix of the PropertyModifiers to keep null values, instead of replacing with an empty string, there is no longer an option to replace a null value with a string. 

So @dgoetz helped me create a new PropertyModifier to be used instead. A test was successful. 

Ref/NC/750293


bobapple:
@cla-bot check

bobapple:
ref/NC/750352
ref/IP/39027

Thomas-Gelf:
Merged, thank you

PR #2528, opened 2022-04-27
Fix restoring director job from snapshot for import source and sync rule

raviks789:
Import source and sync rule ids are added to to director job settings before import as they are not present in the snapshot.

[ref/IP/38986](https://rt.icinga.com/Ticket/Display.html?id=38986&results=ba22c23f42af31dc73f220a432cf3994)


Thomas-Gelf:
Thank you Ravi, has been merged with a small modification (unsetting source/rule after setting the ID on restore)

PR #2530, opened 2022-04-28
DependencyChecker: Use `Module::getRequiredModules()` instead of deprecated method

sukhwinder33445:


Thomas-Gelf:
May I ask what makes it so urgent to forcefully break compatibility with Web v2.8?

nilmerg:
:roll_eyes: The method might be removed in a future update, just keep it as reminder open, or not. :man_shrugging:

Thomas-Gelf:
Fortunately now we know, that keeping it around for 1-2 upcoming versions will make our lives easier ;-)

nilmerg:
Obsoleted by #3065

PR #2534, opened 2022-05-13
Do not use deprecated web functions

nilmerg:
Compatibility with earlier versions than 2.11 is still the case. The commits contain further explanations.


Thomas-Gelf:
Merged, thank you

PR #2544, opened 2022-06-02
Handle cases when gipfl/linux-health returns false instead of memory …

trukonsult:
The module gipfl/linux-health, which Directory uses to read memory usage for the background daemon, returns false if the system doesn't have a /proc-filesystem. This causes the WebUI to fail with an error. This patch allows the WebUI to work and instead shows "n/a" for memory usage in such case.


bobapple:
@cla-bot check

Thomas-Gelf:
Merged, thank you!

PR #2545, opened 2022-06-07
Avoid passing `null` to `strlen` in SyncJob::addSettingsFormFields()

raviks789:


library/Director/Job/SyncJob.php
```diff
@@ -103,7 +103,7 @@ public static function addSettingsFormFields(QuickForm $form)
             )
         ));
 
-        if (! strlen($form->getSentOrObjectValue('job_name'))) {
+        if (! strlen($form->getSentOrObjectValue('job_name') ?? '')) {
```
Thomas-Gelf, line 106: Thanks for this PR, coalesce operator is only allowed since PHP 7.0. Requirements will be raised, but as of this writing, Director still supports 5.6

Thomas-Gelf (changes requested):
Please fix the coalesce operator, then this can be merged

raviks789:
> Please fix the coalesce operator, then this can be merged

Done

Thomas-Gelf:
Merged, thank you!

Issue #2474, opened 2022-02-09
Accessing nested values in import is non-ideal


icanhazpython:
So I am importing via a JSON source. For some record "columns", nested objects are returned and I can access them within the import source and sync rule with dot-notation. Great! However, with my import source, these items can occasionally come back without any object at all, or as *null*. When this occurs, the below-linked error is triggered. I can't seem to figure a way around this other than building a custom importer. I feel that it would be more beneficial to the community however if this functionality was "fixed" so that it handled this situation more gracefully. Here are two possible proposals for fixing this:
* Add a "filter" expression option to ALL import source modifiers (like we have on the sync rules). That way I could simply name the field in the filter, and it would only attempt to access the nested property if the field is not null. 
* Add an option to ALL import source modifiers that allows the user to decide how any error that occurs on a specific field is handled - the "decode a JSON string" import source modifier has an option like this, but I would propose having this option available universally on all modifier types, and additionally adding an option to set a default value other than null. 

I think either of the above options (or even both!) would really benefit other users. If there is some other workaround or option I missed, please let me know!

https://github.com/Icinga/icingaweb2-module-director/blob/302c9b645d901567df5433657487701627396d0d/library/Director/Import/SyncUtils.php#L97

icanhazpython:
Upon further review, I realize that neither of my proposed fixes above would work, as the value access is attempted before any import source modifiers are applied. 

I've submitted a PR with the simplest fix first: simply return null if the value is not a nested object. 
https://github.com/Icinga/icingaweb2-module-director/pull/2584

Another idea I have to fix this:
Create an import source modifier that can extract the nested value, with choices for how to handle an error. The extracted value can then be placed into another variable for further processing if need be.

icanhazpython:
I submitted a second PR for my second fix option: https://github.com/Icinga/icingaweb2-module-director/pull/2585
This fix adds an import source modifier to access nested object properties with configurable error condition

Thomas-Gelf:
@icanhazpython: I have no better solution right now, this would require a new config flag at sync property level. Please have a look at the **Combine** Property Modifier, it should provide what your second pull request does

Thomas-Gelf:
Didn't like the idea of having a flag for this. I changed the default behavior, so for your use-case it should behave as expected out of the box. The following variants are now possible when accessing `app.name`:

* `{ "app": null }`: should ship`null`
* `{ "no app": "something" }`: should ship `null`
[...]

PR #2584, opened 2022-08-14
Accessing nested values in import is non-ideal: fix option 1

icanhazpython:
Fix option 1 for https://github.com/Icinga/icingaweb2-module-director/issues/2474

### DIFF
--- a/library/Director/Import/SyncUtils.php
+++ b/library/Director/Import/SyncUtils.php
@@ -93,11 +93,7 @@ class SyncUtils
             }
 
             if (! is_object($row->$main)) {
-                throw new InvalidArgumentException(sprintf(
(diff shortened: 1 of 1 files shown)


icanhazpython:
CLA signed

bobapple:
@cla-bot check

Thomas-Gelf:
Thank you for your pull request! I didn't merge this, as IMHO it SHOULD fail in case of conflicting types, like `app` being a string or an array (instead of an object) when accessing `app.name`:

* `{ "app": null }`: should ship`null`
* `{ "no app": "something" }`: should ship `null`
* `{ "app": "a string" }`: should fail
* `{ "app": false }`: should fail
* `{ "app": { "name": "app" }`: should ship `"app"`

I'll push a related fix.

PR #2585, opened 2022-08-15
Accessing nested values in import is non-ideal: fix option 2

icanhazpython:
Fix option 2 for https://github.com/Icinga/icingaweb2-module-director/issues/2474


icanhazpython:
Cla has been signed already

bobapple:
@cla-bot check

Issue #2497, opened 2022-03-18
Modification of multiple individual services not possible after upgrade to version 1.9.0


akqopensystems:
## Expected Behavior
In Version 1.8.x and befor it was possible to search for individual service checks and select multiple services. Then you was able to modify or delete the selected services. 

## Current Behavior
After upgrading to version 1.9.0 if we selcet multiple services, there is just only one of them selected for modification. So there is no chance to modify multiple services at once.

## Possible Solution
Maybe there is a issue between uuid and service names?

## Steps to Reproduce (for bugs)
Director -> Services -> single Service -> search for service -> select services on different hosts - on new tab "multiple objects" showing just one item for modification

![example](https://user-images.githubusercontent.com/23358770/159003717-54ca31be-17d2-4e55-ab0e-104e0925a91f.PNG)

## Your Environment
* Director version (System - About): 1.9.0
* Icinga Web 2 version and modules (System - About): 2.9.6
* Icinga 2 version (`icinga2 --version`): 2.12.6
* Operating System and version: RHEL 7.9
* Webserver, PHP versions: 7.3.29

Thomas-Gelf:
I haven't been able to reproduce this with v1.9.1, can you please give it a try?

akqopensystems:
Dear Thomas,
we have implemented the director in Version 1.9.1 and restarted the icinga-director.service, but the problem still exists. If we want to modify or delete multi-selected individual service checks with the same name only one of them is editable.

moreamazingnick:
still broken in 1.10.0

moreamazingnick:
still broken in 1.10.2

Tqnsls:
> I haven't been able to reproduce this with v1.9.1, can you please give it a try?

@Thomas-Gelf  we can reproduce this bug.
Also the older "/icingaweb2/director/services/edit?((id=146)|(id=179)|(id=178))" does not work anymore, it creates an "icinga_service has a multicolumn key, array required".
Please have a look at this.

Tqnsls:
@akqopensystems  maybe change the topic of the issue to "[...] multiple individual services with the same name[...]", to make it more obvious.
I think its important to highlight that the problem does only exist with the same service_names, not when service_name is different.

Thomas-Gelf:
This has been fixed

PR #2594, opened 2022-08-31
Fix multi-renaming-issue refs #2497

moreamazingnick:
$objects[$object->getObjectName()] overwrites services with the same name but doesn't consider the host, so only the last service gets edited.

Since we load the object by uuid we can use the uuid to add the object to the objects array

Best Regards
Nicolas

### DIFF
--- a/application/forms/IcingaMultiEditForm.php
+++ b/application/forms/IcingaMultiEditForm.php
@@ -248,10 +248,15 @@ class IcingaMultiEditForm extends DirectorObjectForm
             $key = 'vars.' . substr($key, 4);
         }
 
-        foreach ($this->objects as $name => $object) {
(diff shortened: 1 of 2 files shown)


Thomas-Gelf:
Good catch, thank you! There is only one issue with the proposed solution:

grafik

Objects keys are being used to show a brief summary regarding configured variants. This now looks quite ugly. You can address it in `IcingaMultiEditForm`, starting from line 250. Instead of using `$name`, you could use `$object->getObjectName()` and optionally combine it with the host name, in case it is a service.

moreamazingnick:
ok, now it's not working

moreamazingnick:
breaks to much, new attempt=> #2599

PR #2626, opened 2022-10-04
Reset Limit count and offset for `$table` in  HostController::servicesAction

raviks789:
The limit in ObjectsTable::prepareQuery() limits the number of services shown in HostController::servicesAction. But this limit is required for pagination in ServicesController. Hence, reset the limit when this query is used in HostController::servicesAction().

[ref/IP/42836](https://rt.icinga.com/Ticket/Display.html?id=42836)

### DIFF
--- a/application/controllers/HostController.php
+++ b/application/controllers/HostController.php
@@ -211,7 +211,8 @@ class HostController extends ObjectController
             ->setAuth($this->Auth())
             ->setHost($host)
             ->setBranch($branch)
-            ->setTitle($this->translate('Individual Service objects'));
+            ->setTitle($this->translate('Individual Service objects'))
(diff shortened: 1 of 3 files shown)


Thomas-Gelf (changes requested):
Resetting those properties is correct, but please use the given constants to do so:

```php
<?php

        $query = $table->getQuery();
        $query->reset($query::LIMIT_OFFSET);
        $query->reset($query::LIMIT_COUNT);
        if (count($table)) {
            $content->add($table);
        }
```

You can also add a `removeLimit()` method to `ObjectsTable` if you want, to avoid duplicate logic.

Thanks,
Thomas

Thomas-Gelf:
Merged, thank you!

PR #2698, opened 2023-01-24
added option to check a single import source or sync rule by name

log1-c:
Did a (possibly) quick and dirty implementation of a new `--name` parameter to make it possible to check a single import source or sync rule with the cli command.

### DIFF
--- a/application/clicommands/HealthCommand.php
+++ b/application/clicommands/HealthCommand.php
@@ -27,7 +27,7 @@ class HealthCommand extends Command
      *
      *   --check <name>     Run only a specific set of checks
      *                      valid names: config, sync, import, jobs, deployment
-	 *   --name <name>		name of a single Import Source or Sync Rule
+     *   --name <name>	    Name of a single Import Source or Sync Rule
      *   --db <name>        Use a specific Icinga Web DB resource
      *   --watch <seconds>  Refresh every <second>. For interactive use only
      */
@@ -46,13 +46,13 @@ class HealthCommand extends Command
 
 
         if ($name = $this->params->get('check')) {
-			if (is_null($check_name)){
-				$check = $health->getCheck($name);
-				echo PluginOutputBeautifier::beautify($check->getOutput(), $this->screen);
-			} else {
-				$check = $health->getCheck($name, $check_name);
-				echo PluginOutputBeautifier::beautify($check->getOutput(), $this->screen);
-			}
+	    if (is_null($check_name)){
+		$check = $health->getCheck($name);
+		echo PluginOutputBeautifier::beautify($check->getOutput(), $this->screen);
+	    } else {
+		$check = $health->getCheck($name, $check_name);
+		echo PluginOutputBeautifier::beautify($check->getOutput(), $this->screen);
+	    }
 
             exit($check->getState()->getNumeric());
         } else {
--- a/application/clicommands/HealthCommand.php
+++ b/application/clicommands/HealthCommand.php
@@ -27,6 +27,7 @@ class HealthCommand extends Command
      *
      *   --check <name>     Run only a specific set of checks
      *                      valid names: config, sync, import, jobs, deployment
+	 *   --name <name>		name of a single Import Source or Sync Rule
      *   --db <name>        Use a specific Icinga Web DB resource
      *   --watch <seconds>  Refresh every <second>. For interactive use only
      */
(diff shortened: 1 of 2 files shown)


log1-c:
Issue birthday is nearing, so I'd like to ask for a current state.
Is there a plan to implement this (or something similar)?

PR Works:
image

boldly pinging @lippserd @Thomas-Gelf

lippserd:
@log1-c PR will be reviewed next week.

application/clicommands/HealthCommand.php
```diff
@@ -37,9 +38,21 @@ public function checkAction()
             $health->setDbResourceName($name);
         }
 
+        if ($this->params->get('name')){
```
lippserd, line 41: This can be simplified to the following:
```php
$checkName = $this->params->get('name');
```

`Params::get()` returns `null` by default if the argument is not specified.

Please also note the change from `$check_name` to `$checkName`. We use camelCase instead of snake_case for class, method, property, variable names, etc.

application/clicommands/HealthCommand.php
```diff
@@ -37,9 +38,21 @@ public function checkAction()
...
+            $check_name = $this->params->get('name');
+        } else { 
+			$check_name = null; 
+		}
+
+
         if ($name = $this->params->get('check')) {
-            $check = $health->getCheck($name);
```
lippserd, line 41: The `echo ...` line remains unchanged, so that it can be used after the `if` and does not have to be repeated. Also, you almost never need to use `is_null()`. We actually only use `=== null` or `!== null`. Additionally, I would start `if`s with what makes them special, which here means that something is different if `$checkName` is actually set, i.e.

```php
if ($checkName !== null) {
    $check = $health->getCheck($name, $checkName);
} else {
    $check = $health->getCheck($name);
}

echo ...
exit(...

File: library/Director/Health.php [hunk 3]
```diff
@@ -36,11 +36,17 @@ public function setDbResourceName($name)
         return $this;
     }
 
-    public function getCheck($name)
+    public function getCheck($name,$check_name = null)
```
lippserd, line 39: There is a space missing after the comma and the variable has to be renamed.

library/Director/Health.php
```diff
@@ -36,11 +36,17 @@ public function setDbResourceName($name)
...
     {
         if (array_key_exists($name, $this->checks)) {
-            $func = $this->checks[$name];
```
lippserd, line 42: Leave that line unchanged. It's the same in both `if` branches.

library/Director/Health.php
```diff
@@ -36,11 +36,17 @@ public function setDbResourceName($name)
...
-            $check = $this->$func();
+            if(!$check_name){
```
lippserd, line 42: A strict `null` check is absolutely necessary if you really want to compare with the specified default value of `null`.  I would also swap the `if` branches again,  i.e. `if ($checkName !== null) { ...`.

library/Director/Health.php
```diff
@@ -155,6 +161,7 @@ public function checkSyncRules()
         foreach ($rules as $rule) {
             $state = $rule->get('sync_state');
             $name = $rule->get('rule_name');
+            if( isset($check_name) && $name !== $check_name){ continue; }
```
lippserd, line 164: Again, use strict `null` checking, i.e. `if ($checkName !== null && ...)`.

lippserd (changes requested):
I think it makes sense to add an option to filter check targets by name. Since jobs also have a name, we could also support the filter here. With the current implementation, the combination of `--name` with `--check deployment` or `--check jobs` would fail in my opinion, as the corresponding filter parameter is missing for the functions. Anyway, I would like to have an implementation that provides roughly the following:

* Support for multiple specification of `--name`. Wildcards should also be supported.
* Analogous support for the specification of `--exclude-name`.
* Trigger an error if `--check deployment` or `--check config` and `--*name are specified`.
* Support filter arguments without having to specify `--check` so that it applies to sync, import and jobs.
* Create a filter object from the specified filter arguments.
* Apply the filter in the corresponding health check functions.

All in all, these changes are not trivial, so I would keep them to a minimum so that we get something done here:

* Raise an error when `--name` is specified with `--check deployment` or `--check config`,  i.e. `$this->fail('--name is not supported with --check ...');`
* Support filtering of jobs.
* Fix code style and indentation 😉.

@log1-c if you need any help with the changes requested, please let me know.

log1-c:
@lippserd 
Thanks for the feedback.

I've done the requested changes regarding the variable naming and the if clauses.
I also added a basic check to raise an error if `--name` is used with `--check deployment` or `--check config`.
I have no idea what the whole filter stuff means or even could look like, sorry.

Issue #2048, opened 2019-12-11
Servicegroup creation error 503 on large environment


lucafwp:
## Describe the bug
Adding a new servicegroup, after clicking on Add and have waited more than 30 seconds, you get ad error "Service Unavailable":

After this error the servicegroup is created correctly anyway.

## To Reproduce

1. Set up an environment with about 90000 services and more than 10000 hosts
2. Create a new servicegroup

## Expected behavior
The servicegroup should be created straight forward without performing a lot of, apparently, useless query on db.

## Screenshots
![image](https://user-images.githubusercontent.com/36190810/70601960-97712900-1bf3-11ea-8d14-09564c367fd9.png)

## Your Environment

* Icinga Web 2 version and modules (System - About): 2.7.3
* Web browser used: Chrome 78.0.3904.108
* Icinga 2 version used (`icinga2 --version`): 2.10.5-1
* PHP version used (`php --version`): 7.1.8
* Server operating system and version: CentOS 7.6.1810

Thomas-Gelf:
I tried it right now on a system with 36k Hosts and 200k Services. Took about five seconds and succeeded. However,  only about 5k of those Services are single ones, the vast majority is being generated by Apply Rules and (applied) Service Sets.

Are you sure about your assumption regarding those DB queries? What kind of queries are you experiencing while this is running?

I see one thing that's going wrong: it loads all Services to pre-calculate group membership for applied Service Groups - even when that Group isn't applied at all. This "should" waste memory and CPU resources, but (in theory) it should fire only a very few fast queries.

The whole behavior here will change in future, as we'll shift these kind of calculations to the background daemon. Still, I'd love to see this fixed also for the current code-base. 10k hosts shouldn't be an issue at all, neither should 90k services. More insight and details would therefore be highly appreciated

As you're running CentOS 7 I guess you're running an old MySQL/MariaDB version? What InnoDB-related settings did you apply? Is the system under heavy load?

lucafwp:
Hi Thomas, we are running CentOS 7 with MariaDB 5.5.60.
We discovered some issues with innodb settings not accepted by system.
These are the settings we would like to use:

large-pages = true
innodb_buffer_pool_size                 = 32GB
innodb_log_buffer_size                  = 4M
innodb_additional_mem_pool_size         = 20M
innodb_thread_concurrency               = 0  

I'll report you back as soon as those settings are validated.

lucafwp:
[...]

PR #2702, opened 2023-02-08
GroupMembershipResolver: do not perform a recheck when dealing with hostgroups

maximecaruchet:
Hello,

I know this fix is probably dirty, but I will be happy to amend this PR and make changes.

The point of this PR is an attempt to fix an issue we are currently facing with Icinga director: when adding a hostgroup Director performs a recheck of all hosts declared within Icinga, which is not scalable at all.

We have thousands of hosts, and creating one new hostgroup through director can take up to 10 minutes now.

After testing on our infrastructure, I still don't really get the point of this all recheck thing, but what I did get is that when creating a host we indeed update some mappings in DB, so even though I don't get the final aim of this, I get that some things are Updated in DB.

However, when creating/deleting hostgroups, I never managed to end up in a case where this whole recheck thing is relevant: the diff always ends up being 0 after the recheck so no mapping needs to be updated.

I don't see the point of iterating over all hosts when adding a new hostgroup (when we add a host we just recheck the host itself but not other hosts). So I just tried to ignore this recheck when dealing with hostgroups.

Again, I'll happily make changes if I can understand what's at stake if I change the code the way I changed it and if this can raise some issues.

### DIFF
--- a/library/Director/Objects/GroupMembershipResolver.php
+++ b/library/Director/Objects/GroupMembershipResolver.php
@@ -23,6 +23,9 @@ abstract class GroupMembershipResolver
     /** @var string Object type, 'host', 'service', 'user' or similar */
     protected $type;
 
+    /** @var string */
+    protected $resolverForType = '';
+
     /** @var array */
     protected $existingMappings;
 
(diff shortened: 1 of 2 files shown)


maximecaruchet:
CLA signed

bobapple:
@cla-bot check

Thomas-Gelf:
Hi @maximecaruchet, thank you for you contribution! Rechecking all Hosts is required whenever a Hostgroup assign filter is created/modified. The issue should have been fixed with #2048. In case it doesn't, please let me know.

Thanks,
Thomas

PR #2743, opened 2023-05-04
fix wrong example in rest-api docs

rndmh3ro:
the example uses wrong parameters to get configs and activities, according to the code: https://github.com/Icinga/icingaweb2-module-director/blob/master/application/controllers/ConfigController.php#L144


Thomas-Gelf:
Merged, thank your

Issue #2286, opened 2021-02-12
Inherited Arguments not shown correctly 


Wintermute2k6:
## Expected Behavior
Fields and Arguments should be shown correctly in the Director with/from inherited templates

## Current Behavior
Fields are shown correctly but Arguments are not shown at all

## Possible Solution
implement rendering for the arguments in the director view for inherited objects. 

## Steps to Reproduce (for bugs)
1. Create a template which includes fields and arguments like a notification command plugin template.
2. Add some fields and arguments.
3. Create a new template or command which inherits this previous template.
4. Go to the Arguments Tab in the new template to see already inherited arguments .. and they are not shown.

## Your Environment
* Icinga Web 2 version and modules (System - About):
```
director       1.8.0     enabled   Director - Config tool for Icinga 2
incubator      0.6.0     enabled   Incubator provides bleeding-edge libraries
ipl            v0.5.0    enabled   The Icinga PHP library
monitoring     2.8.2     enabled   Icinga monitoring module
reactbundle    0.8.0     enabled   ReactPHP-based 3rd party libraries
```
* Icinga 2 version (`icinga2 --version`): 2.12.2-1
* Operating System and version: Centos 7
* Webserver, PHP versions:
```
httpd.x86_64                          2.4.37-21.module_el8.2.0+494+1df74eae      @AppStream
icingaweb2-vendor-lessphp.noarch      2.8.2-1.el8.icinga                         @icinga-stable-release
php.x86_64                            7.2.24-1.module_el8.2.0+313+b04d0a66       @AppStream            
php-Icinga.noarch                     2.8.2-1.el8.icinga                         @icinga-stable-release
php-cli.x86_64                        7.2.24-1.module_el8.2.0+313+b04d0a66       @AppStream            
php-common.x86_64                     7.2.24-1.module_el8.2.0+313+b04d0a66       @AppStream            
php-fpm.x86_64                        7.2.24-1.module_el8.2.0+313+b04d0a66       @AppStream            
php-gd.x86_64                         7.2.24-1.module_el8.2.0+313+b04d0a66       @AppStream            
php-intl.x86_64                       7.2.24-1.module_el8.2.0+313+b04d0a66       @AppStream            
php-json.x86_64                       7.2.24-1.module_el8.2.0+313+b04d0a66       @AppStream            
php-ldap.x86_64                       7.2.24-1.module_el8.2.0+313+b04d0a66       @AppStream            
[...]

PR #2764, opened 2023-06-12
fix/notification-command-inherited-fields-missing refs #2286

moreamazingnick:
Icinga director has different relations for Checks and Notifications but references the same db-table
for checks: check_command
for notifications: command

IcingaObjectfield loader only checks relation check_command


moreamazingnick:
seems to be more complicated since the var resolver doesnt involve inheritance from a notification command. but the inheritance works

moreamazingnick:
If we patch  protected function eventuallyGetResolvedCommandVar(IcingaObject $object, $varName) ind DIrectorDatafield the resolver works works fine.

@Thomas-Gelf  Please tell me if I should prepare another merge request, or if this inheritance behaviour is not a bug but a feature

lippserd:
@moreamazingnick can you please reset your branch to master and reapply your change and then do a force push?

@raviks789 please check this PR.

raviks789:
@lippserd I am not sure I understand the issue and why this change is required. It would nice if there are screenshots of the issue or steps to reproduce the issue.

raviks789:
> Icinga director has different relations for Checks and Notifications but references the same db-table for checks: check_command for notifications: command
> 
> IcingaObjectfield loader only checks relation check_command

I think there is some confusion here. `NotificationCommand` object is not the same as `Notification` object.  Notification commands help to configure how the notifications are sent. Please refer the documentation on [notification commands](https://icinga.com/docs/icinga-2/latest/doc/03-monitoring-basics/#notification-commands) and [notifications](https://icinga.com/docs/icinga-2/latest/doc/03-monitoring-basics/#notifications) for more detail.

When the template created for `NotificationCommand`,  is used to create `NotificationCommand` object then the custom vars are properly inherited by that command. You can then use this command in the `Notification` object.

To create the notification command object. You have to navigate to `Commands` section and add a new command of type `Notification Plugin Command` and import the command template you have created.

Hence, this change which is trying to inherit the custom vars of the notification command template in a notification object is not necessary.

moreamazingnick:
If I create a host or service template and import a command the template shows me all the fields the command has to offer.
since the has relation searches for a "check_command" releation this does not work in the notification context which uses a command relation.

best regards
Nicolas

Issue #2694, opened 2023-01-04
Get PHP error all the time in Icinga Director Deplyments


github6043:
Hi all, after deploying changes in Director, I get the following error all the time (since around 2 Months)

Jan 04 11:17:22 advsrmon122 icingadirector[888]: PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in /usr/share/icingaweb2/modules/director/library/Director>
Jan 04 11:17:22 advsrmon122 icingadirector[888]: Deprecated: Using ${var} in strings is deprecated, use {$var} instead in /usr/share/icingaweb2/modules/director/library/Director/Reso>
Jan 04 11:17:22 advsrmon122 icingadirector[888]: PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in /usr/share/icingaweb2/modules/director/library/Director>
Jan 04 11:17:22 advsrmon122 icingadirector[888]: Deprecated: Using ${var} in strings is deprecated, use {$var} instead in /usr/share/icingaweb2/modules/director/library/Director/Reso>
Jan 04 11:17:22 advsrmon122 icingadirector[888]: PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in /usr/share/icingaweb2/modules/director/library/Director>
Jan 04 11:17:22 advsrmon122 icingadirector[888]: Deprecated: Using ${var} in strings is deprecated, use {$var} instead in /usr/share/icingaweb2/modules/director/library/Director/Reso>
Jan 04 11:17:22 advsrmon122 icingadirector[888]: PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in /usr/share/icingaweb2/modules/director/library/Director>
Jan 04 11:17:22 advsrmon122 icingadirector[888]: Deprecated: Using ${var} in strings is deprecated, use {$var} instead in /usr/share/icingaweb2/modules/director/library/Director/Reso>

I use 

![image](https://user-images.githubusercontent.com/40850921/210559496-84c99f9e-498e-4de8-9f73-f621b4fdf62c.png)

jwilliams-shadowsoft:
Although the message is annoying, and the string interpolation should be fixed in the Director code, that form of string interpolation is still supported through PHP 8. It will become an exception in PHP 9, so the code has to be fixed before then.

[...]

PR #2777, opened 2023-07-19
PHP 8.2: ${var} deprecated in favor of {$var}

EmTeedee:
See: https://www.php.net/manual/en/migration82.deprecated.php#migration82.deprecated.core.dollar-brace-interpolation

Fixes: #2694


Thomas-Gelf:
Thank you, well done!

Issue #2273, opened 2021-01-27
Allow nesting of HostGroups


seferovic:
## Expected Behavior
Icinga allows [nesting of HostGroup objects ](https://icinga.com/docs/icinga-2/latest/doc/09-object-types/#hostgroup) by specifying an array of HostGroups. When specified, the newly created HostGroup will become a member of those groups. This can be used for creation of some top level HostGroups.

## Current Behavior
It is not possible to specify other HostGroups when creating a new HostGroup object.

## Your Environment
* Director version: 1.8.0
* Icinga Web 2 version and modules: 2.8.2
* Icinga 2 version: 2.11
* Operating System and version: RHEL8
* Webserver, PHP versions: Apache 2.4, PHP7

Issue #824, opened 2017-03-01
hostgroup assignments inherited from hosttemplates overwrite each other depending on inheritance sequence


bcogel:
We have defined a tree of hosttemplates to which we assign the services that are needed for kind of  host. To some of those templates we have assigned also hostgroups.

Example: Our AFS database servers and fileservers

Hosttemplates
  |
  SERVICE (dummy template)
    |
    afs     (Import: SERVICE, Services: BOS)
      |
      afsdb (Import: afs, Groups: SYS AFSDB, Services: (BOS from afs) BUSERVER, KASERVER, PTSERVER, VLSERVER)
      |
      afsfs (Import: afs, Groups: SYS AFSFS, Services: (BOS from afs) AFS Space, RXDEBUG)

Preview gives:

template Host "SERVICE" {
}

template Host "afsdb" {
    import "afs"

    groups = [ "SYS AFSDB" ]
}

template Host "afsfs" {
    import "afs"

    groups = [ "SYS AFSFS" ]
}

object HostGroup "SYS AFSDB" {
    display_name = "SYS AFS DB Server"
}

object HostGroup "SYS AFSFS" {
    display_name = "SYS AFS Fileserver"
}

We have some hosts that are both: AFS database server and fileserver

Host afs.thp.uni-koeln.de 

object Host "afs.thp.uni-koeln.de" {
    import "linux-rhel-sys"
    import "afsdb"
    import "afsfs"

    ...
}

Problem: 
Depending on which template comes last in the list of imports, the host is assigned only to the last inherited hostgroup.

Sequence "afsdb afsfs" results in hostgroup "SYS AFSFS", Sequence "afsfs afsdb" results in hostgroup "SYS AFSDB".

To assign hosts like this to more than one hostgroup, I have to edit the group entry of each host manually. Same seems to happen with servicegroups.

In nagios/icinga 1.x it was possible to extend the list of hostgroups by using something like "hostgroups=+afsdb".

dgoetz:
This is also happening for groups at objects or I assume for all attributes of type array.

I would like this to work in the way:
* It shows inherited values
* It allows to override and/or add by user decision additional values

bcogel:
What will happen to attributes inherited from different templates? Do they build up a new array by default that is shown to the user?  Or can I select which one to use? I'm trying to imagine how you want to realize this. 
[...]

Issue #636, opened 2016-11-30
[dev.icinga.com #13375] Append Groups in Templates


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/13375

**Created by leeclemens on 2016-11-30 22:43:11 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2017-01-04 12:54:41 +00:00 (in Redmine)_

---

Append the groups array values, rather than re-assigning (similar to https://dev.icinga.com/issues/12941).

Use Case is having multiple Host Templates with Groups, only the group(s) assigned to the lowest level template are assigned to the Host object.

**Attachments**

* [0001-Append-groups-array.patch](https://dev.icinga.com/attachments/download/3861/0001-Append-groups-array.patch) leeclemens - _2016-11-30 22:46:22 +00:00_

---

**Relations**:

* relates [#13375](https://dev.icinga.com/issues/13375)

thille-IM:
I ran into the problem too, when I used templates to add hosts to groups (and other things). The hosts were only added to the group(s) defined in the last imported template.

The solution mentioned by tobiasvdk, adding a boolean flag (checkbox) to decide wether to extend or replace an array is IMHO the best in terms of ease of use and probably easiest to implement.

Issue #344, opened 2016-07-07
[dev.icinga.com #12122] Template inheritance problem


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12122

**Created by spillerm on 2016-07-07 07:41:39 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-12-06 23:51:05 +00:00 (in Redmine)_

---

Hi,

I noticed the following: in Director, I created a host template "generic-cisco", preview of this template is:

@
template Host "generic-cisco" {
import "generic-switch-template"

groups = \[ "corporate", "network" \]
vars.is\_cisco = "true"
}
@

Now I would like to add a switch, so I create an object which imports generic-cisco; the field "groups" remains empty, and even when I'm looking at the "resolved" preview, my group assignment does not appear.
So at this point, inheritance does not seem to work.

Regards,
Marianne

---

**Relations**:

* relates [#13375](https://dev.icinga.com/issues/13375)

lesinigo:
We are seeing many organizations that would really like to have this feature implemented, especially those coming from `$anothermonitoringsoftware` that always had it. Even without the "migrate from older software" issue, using multiple template layers that add multiple host group layers (eg. from the more generic to the more specific ones) is a really common (and IMHO good) practice.

The proposal by @Thomas-Gelf in https://github.com/Icinga/icingaweb2-module-director/issues/344#issuecomment-273105151  "_element showing inherited array elements somehow within the same elements allowing one to delete some, add others and/or decide to override all of them_" sounds like a really good solution, but the linked issue #636 seems to have stalled some years ago (!).

If it could be more feasible, I think the boolean flag proposed by @tobiasvdk in https://github.com/Icinga/icingaweb2-module-director/issues/636#issuecomment-273106508 could be more than enough for many people. I actually tend to advise against "exclusions" in this kind of stuff because it has a tendency to result in complex spaghetti configurations in the long term, so I usually suggest to go "append only" and such a flag would be perfect for that.

Hope this issue will get some attention in the not too distant future :-)

afeefghannam89:
@Thomas-Gelf ref #734257

sol1-matt:
[...]

PR #2782, opened 2023-08-07
Add global element for default group behavior.

sol1-matt:
Add Global Director Settings for group behavior so it can be changed from the current default of assignment (=) to append (+=). 

- Add global element for default group behavior.
- Add default value 'assign' for default group behavior which matches group behavior . 
- Add logic to IcingaObjectGroups so the class initializes the with the global group behavior and sets the group operator based on the class operator. 

Partial Fix for #344, #636, #824, #2273

### DIFF
--- a/application/forms/SettingsForm.php
+++ b/application/forms/SettingsForm.php
@@ -39,6 +39,25 @@ class SettingsForm extends DirectorForm
             'value'  => $settings->getStoredValue('default_global_zone')
         ));
 
+        $this->addElement('select', 'default_group_behaviour', array(
+            'label'        => $this->translate('Default group behaviour'),
+            'multiOptions' => $this->eventuallyConfiguredEnum(
+                'default_group_behaviour',
+                array(
+                    'assign' => $this->translate('Assign'),
+                    'add' => $this->translate('Add'),
+                    // Remove (-=) is a group option but global remove isn't useful as all groups would be removed all the time.
+                )
+            ),
+            'description'  => $this->translate(
+                'Icinga Director deploys group objects using the operator assign (=)'
+                . ', which overrides groups set in inherited objects, by default.'
+                . ' This option allows you to adjust the group operator globally to'
+                . ' add (+=) which appends groups from inherited objects.'
+            ),
+            'value'  => $settings->getStoredValue('default_group_behaviour')
+        ));
+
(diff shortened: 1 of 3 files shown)


sol1-matt:
> Please follow instructions at https://icinga.com/company/contributor-agreement to sign the CLA.

CLA has been signed

lippserd:
@cla-bot check

Thomas-Gelf:
Hi @sol1-matt,

most users probably want it this way, but for compatibility reasons this is hard to change. While I do not really like the idea of a global toggle, this of course addresses the problem. From a user perspective it might be irritating, having on Director behaving this way and another Director behaving another way. Even worse for vendors shipping baskets, that might behave differently in different setups.

To complete your variant, there is one important thing missing: object property resolution must work the same way, to address pre-calculated group memberships and related UI features. But even given this, I would prefer a clean solution, which would have to store the desired behaviour with every assignment.

Cheers,
Thomas

sol1-matt:
Hi @Thomas-Gelf,

I did look into individual sync rule properties, it is where the changes in [library/Director/Objects/IcingaObjectGroups.php](https://github.com/Icinga/icingaweb2-module-director/pull/2782/files#diff-9af13124d81430caf399c36a7ce59303819aaddaf25d5326b46d791f88dcb806) came from originally, but had some problems getting that working, from memory the database needs to be modified to accommodate the use of assign, remove and append.  

If the preferred approach here is to implement this at the individual sync rule property level I'll look at what changes are needed to make that happen. 

Thanks, 
Matt

Thomas-Gelf:
@sol1-matt: this is about group inheritance behavior, not about Sync - isn't it?

sol1-matt:
@Thomas-Gelf 

This is about adding the ability for groups to use inheritance. 

But I do a lot of automation so my solutions focused on functionality that can be used in Automation Sync Rules to achieved the desired outcome, or the global setting which affects everything. 

Your comment is a reminder than there is a whole other UI people use to enter things manually so I'll need to account for that as well which will be complicated. The Automation Sync Rule properties are ordered, the current group allocation UI in director not so much. 

Are you sure you don't want to reconsider your stance on a global setting? 
I'm managing multiple director installations, many of which I've added a global group append to. I've found the advantages of having a global setting far outweigh the downsides you've listed above in practical use.

Issue #2788, opened 2023-08-24
FeatureRequest DeploymentHook-function/BeforeStoreIcingaObjectHook


moreamazingnick:
@Thomas-Gelf @lippserd 
I would like to propose a new deploymenthook function to make changes to the files written to disk.

In addition I would like to propose a BeforeStoreIcingaObjectHook, which allows changes to an IcingaObject before written to the Database.

Please reach out to me if any changes are necessary.
I think linuxfabrik could make use of the deploymenthook function as well.

PullRequest will be created soon.

Best Regards 
Nicolas

Issue #2752, opened 2023-05-19
Require confirmation before passwords get deployed to every Icinga2 agents


slalomsk8er:
## Expected Behavior
The director assists in managing a secure Icinga2 configuration by preventing accidental distribution of passwords to all Icinga2 agents.

## Current Behavior
Without knowledge on how variables get written to the Icinga2 configuration and zones especially, it's easy to write a password that is only required on the master or one agent to the whole cluster.

## Possible Solution
Require confirmation and remember the allowed zones if a password needs to be deployed outside of the master.

PR #2790, opened 2023-08-24
Feature/new deployment hook

moreamazingnick:
refs #2788 

best regards 

Nicolas

### DIFF
--- a/library/Director/Core/CoreApi.php
+++ b/library/Director/Core/CoreApi.php
@@ -845,17 +845,19 @@ constants
             // 'username'       => Util::getUsername(),
             // 'module_name'    => $moduleName,
         ));
+        $files = $config->getFileContents();
 
         /** @var DeploymentHook[] $hooks */
         $hooks = Hook::all('director/Deployment');
         foreach ($hooks as $hook) {
             $hook->beforeDeploy($deployment);
+            $files = $hook->beforeDump($files);
         }
 
         $this->assertPackageExists($packageName);
 
         $response = $this->client()->post('config/stages/' . \rawurlencode($packageName), [
-            'files' => $config->getFileContents()
+            'files' => $files
         ]);
 
         $duration = (int) ((microtime(true) - $start) * 1000);
--- a/library/Director/Hook/BeforeStoreIcingaObjectHook.php
+++ b/library/Director/Hook/BeforeStoreIcingaObjectHook.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace Icinga\Module\Director\Hook;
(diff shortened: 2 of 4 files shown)


Thomas-Gelf:
Hi @moreamazingnick,

I have some problems with this PR, and cannot merge it as is. Main problems:

* tweaking config files AFTER their checksums have been calculated and persisted is a very bad idea. Please explain your motivation behind this. If this is about shipping additional configuration, not controlled by the Director, then the ShipConfigFilesHook could be what you're looking for. If you want to have an influence on rendering behaviour, a more specific Hook should be found
* allowing to modify every object after all validation took place also doesn't make me feel comfortable. Of course it gives your implementation super powers, and this sounds promising. But it makes future development and major changes incredibly difficult, as we can no longer trust in our objects being stored as given. Lots of modules might implement black magic, and they might even conflict, depending on execution order.

This all also might result in undesired side-effects related to caching mechanisms, Apply Rule calculation and more. If you want to tackle this, please provide related real-world examples, which led to your requirements. I'm sure we'll find a better solution, to address them.

Cheers,
Thomas

moreamazingnick:
Hi @Thomas-Gelf 

Thanks for the reply.

With these hooks I created something like this in a non prod environment, just a proove of concept:
* Create a Field that has the ability to wrap a passsword. PREENC()
* Since fieldnames are not unique, I don't know, which field is an encrypted field so I need something like this "modify every object"-hook to encrypt everything that matches PREENC(.+?)
* Encrypt the fieldcontent and change it to ENC(xxx)
* Since icinga2 cannot make use of my decryption feature I need to write it to disk unencryted so I need the "tweaking config files AFTER"-hook and replace the ENC(xxx) with the clear text password.

Why I think it is a good idea:
* It provides the possibility to use PREENC(secretxy) in api too.
* no passwords are leaked in any way using director web gui
* passwords are "protected" in director database, backups, dumps
* * since the director writes configfiles in plaintext to the db this has to be done after checksum creation so the secrets are still protected in the database


It sadly moves the problem to the icinga2 infrastructure but the director database is somehow secured.

here a module that makes use of these hooks:
https://github.com/moreamazingnick/directorencrypter

other ideas where the"modify every object"-hook would help:
* it could also be used to warn the user that a password is saved to a template the user can revert the changes before written to disk https://github.com/Icinga/icingaweb2-module-director/issues/2752
* it could trim object names for trailing whitespaces or prevent object_names that make troubles. https://github.com/Icinga/icingaweb2-module-businessprocess/issues/412


Best Regards Nicolas

PR #2309, opened 2021-04-08
Add php action workflow

nilmerg:
Unit tests are still missing

PR #2792, opened 2023-08-29
Add PHPStan and ensure PHP 8.2 support

raviks789:
Dynamic properties are deprecated in PHP 8.2 and must be avoided.


nilmerg (changes requested):
please add phpstan with level 2 to the workflow configuration. Plus a baseline configuration.

nilmerg:
Workflow configuration is available in master since today. And of course, add PHP 8.2 to the matrix.

.github/workflows/php.yml
```diff
@@ -31,7 +31,47 @@ jobs:
           tools: phpcs
 
       - name: Setup dependencies
-        run: composer require -n --no-progress overtrue/phplint
+        run: |
+          composer require -n --no-progress overtrue/phplint
+          git clone --depth 1 https://github.com/Icinga/icingaweb2.git vendor/icingaweb2
+          git clone --depth 1 -b snapshot/nightly https://github.com/Icinga/icinga-php-library.git vendor/icinga-php-library
+          git clone --depth 1 -b snapshot/nightly https://github.com/Icinga/icinga-php-thirdparty.git vendor/icinga-php-thirdparty
+          git clone --depth 1 -b main https://github.com/Icinga/icingaweb2-module-cube.git vendor/cube
```
nilmerg, line 39: please remove -b main, shouldn't be necessary

nilmerg:
Please also take a look at the unit test output on PHP 8.1 and 8.2. There are still some deprecations logged.

raviks789:
> Please also take a look at the unit test output on PHP 8.1 and 8.2. There are still some deprecations logged.

This is because the replaced `ramsey/collection` package loses the PHP 8.1 support patch written in icinga-php-thirdparty. There is an issue gipfl/icinga-bundles#1 opened in gipfl/icinga-bundles repository for this.

nilmerg (approved):

Thomas-Gelf (changes requested):
@raviks789: good work, thank you! Could you please install the incubator version required in our module.info? If parsing the file annoys you, I'm fine with a hard-coded one for our tests, which can be raised once required.

Testing with unreleased dependencies is a good thing, which could be added as an optional/additional workflow. Main priority should be testing what's going to be installed. This then also removes the requirement to track incubator module dependencies in a second place.

Many thanks
Thomas

nilmerg:
The installed incubator version has been introduced by #2309, not here. And this has the same intended effect as the just mentioned change: Testing the *current* state. If you want a stable test workflow, we should add this separately. Though, only for `release/*` branches or similar, there's no reason to run such a test on every push.

Thomas-Gelf:
For the Director testing with library versions it will never run with has absolutely no purpose. Here we're testing the current Director state, and not any intermediary foreign library state, that might or might not be released.

Thomas-Gelf:
Merged, thank you!

PR #2797, opened 2023-09-06
Remove limit count for service set directly added to the host

raviks789:
fixes [ref/IP/47505](https://rt.icinga.com/Ticket/Display.html?id=47505)


Thomas-Gelf:
Merged, thank you!

PR #2801, opened 2023-09-12
Fix editing of multi-selected services with same name

### LINKED ISSUE #2798: Can't edit multiple services with same name
## Expected Behavior

As a user I would like to change some custom properties on a service on multiple hosts.

## Current Behavior

At the moment, if you go to "Icinga Director >> Services >> Single Services" and select multiple services with the same name, only one service is actually selected:
image

- Create 2 hosts: host1, host2
- Create a service: service_test
- Add service_test on both hosts
- Go to "Icinga Director >> Services >> Single Services"
- Select service_test from the two hosts
- Edit some custom properties

## Your Environment

The issue has been reproduces on icingaweb2 demo: https://icinga.com/demo
* Director version (System - About): 1.9.1
* Icinga Web 2 version and modules (System - About): 2.11.1
* Icinga 2 version (`icinga2 --version`): XX
* Operating System and version: XXX
* Webserver, PHP versions: XXX

raviks789:
fixes #2798

### DIFF
--- a/library/Director/Web/Controller/ObjectsController.php
+++ b/library/Director/Web/Controller/ObjectsController.php
@@ -13,6 +13,7 @@ use Icinga\Module\Director\Forms\IcingaMultiEditForm;
 use Icinga\Module\Director\Objects\IcingaCommand;
 use Icinga\Module\Director\Objects\IcingaHost;
 use Icinga\Module\Director\Objects\IcingaObject;
+use Icinga\Module\Director\Objects\IcingaService;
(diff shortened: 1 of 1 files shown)


Thomas-Gelf (approved):

Thomas-Gelf (changes requested):
Key is used for visualization purposes only, so we could use ': ' (trailing space) as a separator. If you check for `instanceof IcingaService` instead of `$type === 'service'`, phpstan should no longer complain

Thomas-Gelf:
Merged, thank you!

PR #2803, opened 2023-09-20
More granular Director permissions

jschlegelmilch:
Hi,
as for our needs we have been changing permission flags for several director objects to manage them better for a company where more restricted access to several options is needed.
Them main task was to change from an admin-only permission to a few different set of new permissions.


Thomas-Gelf:
It's still a draft, but as you already invested quite some time, please let me drop some notes:

* at first sight, this might introduce security issues, as it currently circumvents existing permission checks by replacing `assertPermission()` (which would trigger a SecurityError) with `hasPermission()` (which just tells you, whether a specific permission has been granted
* we cannot merge a PR, that would break existing setups. e.g.: someone who is allowed to delete Hosts right now, must be allowed to do so also in the next release, without being forced to tweak role permissions
* please use phpcs for formatting hints

Cheers,
Thomas

jschlegelmilch:
Hi Thomas,

thanks for already reviewing and dropping some notes.
I will wrok on these points and will send feedback again after im done with all the tasks.
Will it be okay for you, if we squash the commits, for a better general overview for reviewing?

Issue #2410, opened 2021-10-12
Isolation of deployed "secrets" between independent zones


ghost:
Hi!
As in agreement with @lippserd below is the write up of our proposal for the isolated deployment of secrets to separate zones.

---
Let's assume we have a distributed monitoring environment with four satellites.

Furthermore, always two of these satellites monitor the infrastructure of two different, completely independent, real world entities, and as such shall be isolated from each other:
* EntityA is monitored by SatA1(Zone A_Int) and SatA2(Zone A_Cloud)
* EntityB is monitored by SatB1(Zone B_RZ_DE) and SatB2(Zone B_RZ_FI)

When monitoring services which need credentials, an example could be the _SNMP Community_, the correct mechanism to distribute the secrets to their respective icinga2 instances, seems to be assignment via Hosttemplates.

## Current Behavior
When using the director-global zone for these Hosttemplates, they are distributed among the entire monitoring environment, thus _information is leaked_ between the Entities.

Also when assigning a zone to the Hosttemplates, they get deployed _only_ to that specific single zone.

This requires that for having them on each zone of an Entity, they must be duplicated N times for each zone of the Entity.
In our example it's not so bad as N=2, but in reality N can be _dozens_ of zones and satellites.

## Expected Behavior
A way to deploy "secret" custom variables to a limited subset of zones, without the need of duplicated configuration for each zone.

### Proposal: On demand Hosttemplate deployment
Implicitly deploy a host template to a zone, always when it is directly or transitively associated with a Host belonging to this Zone.

Hosttemplates:
* generic-host in director-global zone
* entityA-secrets containing customvar snmp_community=SecretA
* entityA-host-secret imports generic-host && entityA-secret

Hosts:
* host.A_int.example.com imports entityA-host-secret and assigned to Zone A_Int.
* host.B_RZ_DE.example.com imports generic-host and assigned to Zone B_RZ_DE.

What should happen in this case, is that independently of the zone of the last two Hosttemplates:
* the deployment to Zone A_Int should include generic-host via director-global, but also entityA-secrets and entityA-host-secred by transitivity, as Host host.A_int.example.com is assigned to this Zone and inherits from these Hosttemplates.
* the deployment to Zone B_RZ_DE should include only the generic-host Template via director-global Zone.

There are obviously some details left to refine:
* When does a Template get deployed to director-global zone, and when on-demand deployed?
  Maybe a new zone-dropdown entry "on-demand"?
[...]

PR #2804, opened 2023-09-21
IcingaConfig: render host templates to all...

Thomas-Gelf:
...non-global zones, instead of the default global zone

refs #2410


raviks789 (approved):

raviks789:
Merged, thank you

PR #2809, opened 2023-10-09
fix: update filterHostGroupsQuery default argument

### LINKED ISSUE #2164: HostGroup edit fails when director/filter/hostgroups is active
## Expected Behavior
Work as normal editation (at least allow Assign where changes) of given Group to users that has allowed that group by director/filter/hostgroup setting.
Solving that issue may be a partial workarround to issue #2163 

## Current Behavior
An error occures when trying to edit given Host Group:
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'o.object_name' in 'where clause', query was: SELECT h.id FROM icinga_hostgroup AS h WHERE (id = '4') AND (o.object_name IN ('Linux'))

SQLSTATE[42S22]: Column not found: 1054 Unknown column 'o.object_name' in 'where clause'

## Possible Solution
Change function filterHostGroupsQuery default argument to $tableAlias = 'h' instead of current value 'o' at file library/Director/Restriction/HostgroupRestriction.php

- Limit user role to access only some Host Groups in Icinga Director by setting director/filter/hostgroups to correct value/name(s) of existing Host Group(s).
- Logout/Login with that role.
- Try to edit given Host Group with that user, an error above will be displayed.

## Your Environment
* Director version (System - About): 1.7.2 or older (tested also on 1.6.2)
* Icinga Web 2 version and modules (System - About): N/A
* Icinga 2 version (`icinga2 --version`): N/A
* Operating System and version: N/A
* Webserver, PHP versions: N/A

0xliam:
Resolves #2164

Happy to work with the Icinga team on this PR - please let me know if anything needs changing.

When a role restricts users to a particular host group, e.g:

image

The following error is returned when trying to edit any host group:

```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'o.object_name' in 'where clause', query was: SELECT h.id FROM icinga_hostgroup AS h WHERE (id = '324') AND (o.object_name IN ('example'))
```

This PR updates the filterHostGroupsQuery default argument to 'h', as per the fix outlined suggested in #2164.

I've been using this in our production environment and it doesn't appear to have caused any adverse side effects.

### DIFF
--- a/library/Director/Restriction/HostgroupRestriction.php
+++ b/library/Director/Restriction/HostgroupRestriction.php
@@ -131,7 +131,7 @@ class HostgroupRestriction extends ObjectRestriction
      * @param ZfSelect $query
      * @param string $tableAlias
      */
-    protected function filterHostGroupsQuery(ZfSelect $query, $tableAlias = 'o')
+    protected function filterHostGroupsQuery(ZfSelect $query, $tableAlias = 'h')
     {
         if (! $this->isRestricted()) {
             return;


0xliam:
Signed CLA

bobapple:
@cla-bot check

Thomas-Gelf:
Hi @0xliam, thanks for your submission! I fixed it in a slightly different place, where the method call context (directly after the related query) makes it more obvious, why that specific alias fits here.

Issue #2815, opened 2023-10-16
Cannot select data field on service template without a check command


lazyfrosch:
## Expected Behavior

I should be able to select and add any field to a service template, even when no check command is selected. This was working before one of the later releases.

## Current Behavior

![image](https://github.com/Icinga/icingaweb2-module-director/assets/121874/d3f90c91-8fae-4cab-98af-947265123bba)

Error: `There are no data fields available. Please ask an administrator to create such`

Only when a check command is set, the list is usable:

![image](https://github.com/Icinga/icingaweb2-module-director/assets/121874/b32dc161-a749-496f-8e98-9135ef346195)

## Your Environment
* Director version (System - About): 1.11.0
* Icinga Web 2 version and modules (System - About): 2.12.0
* Operating System and version: CentOS 7
* Webserver, PHP versions: 7.3.33

PR #2840, opened 2023-12-05
Fix form field suggestions

### LINKED ISSUE #2826: Problem New custom field in Notification Templates
I have a problem when I try to add a new "custom field" inside a notification templates"

I correctly created a data field categories and added 2 data fields, but when I go to add it to the template notification it gives me an error which I report in the attachment

## Your Environment

* Director version (1.11.0):
* Icinga Web 2 version and modules (2.12.0):
* Icinga 2 version (r2.14.0-1):
* Operating System and version:
* Debian 11.8

Image 003

raviks789:
Other available fields should be suggested when no check command has been chosen for service template.

fixes #2826, #2815

### DIFF
--- a/application/forms/IcingaObjectFieldForm.php
+++ b/application/forms/IcingaObjectFieldForm.php
@@ -53,13 +53,8 @@ class IcingaObjectFieldForm extends DirectorObjectForm
             $command = null;
         }
 
-        if ($command) {
-            $suggestions = $this->fieldSuggestion = new FormFieldSuggestion($command, $this->db->enumDatafields());
-            $fields = $suggestions->getCommandFields();
-        } else {
-            $suggestions = null;
-            $fields = [];
-        }
+        $suggestions = $this->fieldSuggestion = new FormFieldSuggestion($command, $this->db->enumDatafields());
+        $fields = $suggestions->getCommandFields();
 
         $this->addElement('select', 'datafield_id', [
             'label'        => 'Field',
@@ -97,7 +92,7 @@ class IcingaObjectFieldForm extends DirectorObjectForm
                     . ' user puts the focus on this field'
                 ),
                 'ignore'      => true,
-                'value'       => $suggestions ? $suggestions->getDescription($id) : null,
+                'value'       => $command ? $suggestions->getDescription($id) : null,
                 'rows'        => '3',
             ]);
         }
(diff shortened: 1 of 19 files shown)


RincewindsHat:
does not work for notifications for a first test

raviks789:
> does not work for notifications on a first test

As discussed offline, this is a fix for the bug where the data fields are not available for selection when the check commands are not selected (Not a feature to show the custom properties of notification command the notifications form).

bigfreak85:
works and has helped me to get the job done

application/forms/IcingaObjectFieldForm.php
```diff
@@ -167,26 +179,4 @@ public function onSuccess()
         $this->object()->set('var_filter', $this->getValue('var_filter'));
         parent::onSuccess();
     }
-
-    protected static function getFilterFields(IcingaObject $object): array
```
nilmerg, line 171: Why did you remove this? It doesn't seem relevant to the fix.

nilmerg (changes requested):

test/php/library/Director/Form/IcingaObjectFieldFormTest.php
```diff
@@ -0,0 +1,153 @@
...
+
+    /** @var ?IcingaCommand */
+    protected $testIcingaCommand = null;
+
+    /** @var IcingaHost */
+    private $testHost;
+
+    public function setUp(): void
+    {
+        $this->app()->setRequest(new Request());
```
nilmerg, line 35: This is not required if the director's `BaseTestCase` extends Icinga Web's `BaseTestCase`. At least, works for me locally. Remember to call the parent's `setUp` then.

test/php/library/Director/Form/IcingaObjectFieldFormTest.php
```diff
@@ -0,0 +1,153 @@
...
+
+        if ($this->hasDb()) {
+            $db = $this->getDb();
+            $this->tearDown();
```
nilmerg, line 39: Don't call this in `setUp`. I don't even dare to think of the side effects.

test/php/library/Director/Form/IcingaObjectFieldFormTest.php
```diff
@@ -0,0 +1,153 @@
...
+            [
+                'Other available fields' => [
+                    $this->testDatafield->get('id') => "Blah (dataFieldTest)"
+                ]
+            ],
+            $suggestions
+        );
+    }
+
+    public function tearDown(): void
```
nilmerg, line 125: Call the parent once `BaseTestCase` extends `Icinga\Test\BaseTestCase`

nilmerg (changes requested):

nilmerg (changes requested):

nilmerg (approved):

PR #2857, opened 2024-01-12
Support Content-Security-Policy

### LINKED ISSUE #2845: Support strict CSP
Icinga Web 2.12 introduced the ability to enable strict content security policy that the Director does not support yet. Inline JS must be avoided. Inline CSS should be avoided or protected by a nonce.

sukhwinder33445:
resolves #2845


public/css/module.less
```diff
@@ -1833,3 +1872,38 @@ table.table-basket-changes {
...
+    color: green;
+  }
+}
+
+.seviceset-obj-link {
+  float: right;
+  font-weight: normal
+}
+
+//DataController::dictionaryAction, cannot findusage in web
```
raviks789, line 1902: You can remove this comment, I have tested this.

application/forms/AddToBasketForm.php
```diff
@@ -41,21 +38,14 @@ public function setup()
...
-                $count++;
+            if (! $basket || ! $basket->hasObject($type, $name)) {
                 $names[] = $name;
+                $count++;
             }
         }
         $this->addHtmlHint((new HtmlDocument())->add([
             'The following objects will be added: ',
-            $names
+            (array) implode(", ", $names)
```
raviks789, line 48: instead of casting the string to array could you not perform string concatenation?

nilmerg (changes requested):
There's still an onclick attribute defined in `ImportsourceController:183`

nilmerg (approved):

nilmerg:
please rebase

Issue #2827, opened 2023-10-30
Self Service API not working with PHP 8.2.7


seoni-at:
## Expected Behavior
Self service should work

## Current Behavior
Icinga For Windows fails using the Self Service API.

## Possible Solution

This code Block produces deprecated output and prevents parsing of the json.
https://github.com/Icinga/icingaweb2-module-director/blob/aa31b37fd184228d3397c512ae8fc175f2d82d39/application/controllers/SelfServiceController.php#L403-L411
```
<b>Deprecated</b>:  strlen(): Passing null to parameter #1 ($string) of type string is deprecated in <b>/usr/share/icingaweb2/modules/director/application/controllers/SelfServiceController.php</b> on line <b>407</b><br />\n
```

I fixed it in my installation with checking if the $value is null but my php knowledge is not the best so i do not know if there is a better way.
 
```php
    protected function addStringSettingsToParams(Settings $settings, array $keys, array &$params)
    {
        foreach ($keys as $key) {
            $value = $settings->get("self-service/$key");
            if ($value != null) {
                if (strlen($value)) {
                    $params[$key] = $value;
                }
            }
        }
    }
```

## Steps to Reproduce (for bugs)
Install icinga director on Debian 12 bookworm with php 8.2.7 and try to add a host using the self service api.
Using wireshark the response sent to icinga for windows is the following

```
<br />
<b>Deprecated</b>:  strlen(): Passing null to parameter #1 ($string) of type string is deprecated in <b>/usr/share/icingaweb2/modules/director/application/controllers/SelfServiceController.php</b> on line <b>407</b><br />
<br />
<b>Deprecated</b>:  strlen(): Passing null to parameter #1 ($string) of type string is deprecated in <b>/usr/share/icingaweb2/modules/director/application/controllers/SelfServiceController.php</b> on line <b>407</b><br />
{
    "fetch_agent_name": false,
    "fetch_agent_fqdn": true,
    "transform_hostname": "1",
    "flush_api_directory": true,
    "resolve_parent_host": "0",
    "install_framework_service": "0",
    "install_framework_plugins": "0",
    "director_host_object": "{\"address\":\"&ipaddress&\",\"display_name\":\"&hostname.lowerCase&\"}",
    "download_url": "https://packages.icinga.com/windows/",
    "agent_version": "2.14.0",
    "allow_updates": true,
    "agent_listen_port": 5665,
    "install_nsclient": true
} 
```
## Your Environment

[...]

Issue #2819, opened 2023-10-17
Director kickstart Wizard - Migration after Update - PHP Error


Pio84:
## Expected Behavior
run import on Director Kickstart Wizard

## Current Behavior
Director Kickstart Wizard shows 23 times the following error

Deprecated
: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in
/usr/share/icingaweb2/modules/director/library/Director/Objects/IcingaCommand.php
on line
131

## Possible Solution
no idea

## Steps to Reproduce (for bugs)
after updating:
Icinga Web2 from 2.11.4 to 2.12.0
Icinga DB Web from 1.0.2 to 1.1.0
Icinga Director from 1.10.2 to 1.11.0
PHP from 8.1.2 to 8.2.11
php-thirdparty from 0.11.0 to 0.12.0
php-libary from 0.12.0 to 0.13.0

i got the director "migrate" button in the kickstart menu an the error described in ## Current Behavior

## Your Environment
Icinga2 r.2.14.0-1 
Icinga Web2 2.12.0
Icinga DB 1.1.1
Icinga DB Web 1.1.0
Icinga Redis 7.0.12
Icinga Director 1.11.0 
Apache2 2.4.52
PHP 8.2.11
MariaDB 10.8.8
Phyton 3.10.12
php-libary 0.13.0
php-thirdparty 0.12.0
incubator 0.20.0
pdfexport 0.10.2

Pio84:
![2023-12-15 09_00_31-Icinga](https://github.com/Icinga/icingaweb2-module-director/assets/48513287/b72ec93a-dd07-4a66-80c5-f32bb7e61d56)

PR #2858, opened 2024-01-17
Phpstan: Fix deprecated function param types

sukhwinder33445:
fixes:
https://github.com/Icinga/icingaweb2-module-director/issues/2827
https://github.com/Icinga/icingaweb2-module-director/issues/2819

### DIFF
--- a/application/controllers/ConfigController.php
+++ b/application/controllers/ConfigController.php
@@ -370,7 +370,7 @@ class ConfigController extends ActionController
 
         $configs = $db->enumDeployedConfigs();
         foreach (array($leftSum, $rightSum) as $sum) {
-            if (! array_key_exists($sum, $configs)) {
+            if ($sum && ! array_key_exists($sum, $configs)) {
                 $configs[$sum] = substr($sum, 0, 7);
             }
         }
--- a/library/Director/Core/RestApiClient.php
+++ b/library/Director/Core/RestApiClient.php
@@ -206,14 +206,14 @@ class RestApiClient
     }
 
     /**
-     * @return resource
+     * @throws RuntimeException
      */
     protected function curl()
     {
         if ($this->curl === null) {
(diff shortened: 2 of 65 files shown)


library/Director/Core/RestApiClient.php
```diff
@@ -260,13 +260,11 @@ protected function processReadBuffer($offset, $pos)
 
     public function disconnect()
     {
-        if ($this->curl !== null) {
-            if (is_resource($this->curl)) {
-                @curl_close($this->curl);
```
nilmerg, line 265: Is removing the `@` really necessary?

sukhwinder33445: No, but it looks superfluous to me. This function cannot throw an error.

sukhwinder33445:
Note:
 Phpstan failed to detect `ConfigController:373`:  substr(): Passing null to parameter#1 ($string) of type string is deprecated

application/controllers/ConfigController.php
```diff
@@ -370,7 +370,7 @@ public function diffAction()
 
         $configs = $db->enumDeployedConfigs();
         foreach (array($leftSum, $rightSum) as $sum) {
-            if (! array_key_exists($sum, $configs)) {
+            if ($sum && ! array_key_exists($sum, $configs)) {
```
sukhwinder33445, line 373: Note:
 Phpstan failed to detect this:  substr(): Passing null to parameter#1 ($string) of type string is deprecated

This is a mystery to me, because 2 days ago phpstan showed me the error.

library/Director/Core/RestApiClient.php
nilmerg: It cannot? Why? If you're unsure, please restore it.

sukhwinder33445: It's not documented, but yes, will restore it.

nilmerg (approved):

PR #2861, opened 2024-01-19
Remove `gipfl/format` package dependency

### LINKED ISSUE #2847: Remove usage of `gipfl\Format\LocalTimeFormat` in favor of direct use of `intl` also fixing bugs
When using `intl` formatters, we may pass `C` as the locale which is not a valid ICU locale (note that this is not just `gipfl` specific). This was fine until a recent change in [PHP](https://github.com/php/php-src/issues/12561), which now throws an exception in PHP >= 8.1.25. Since patching `gipfl` is not an option, we will simply replace it with using `intl` directly. Note that it is important not to use `setlocale(..., 0)` to pass the locale to `intl`, as this could pass `C`, which is invalid. I suggest just using `null` as this will automatically retrieve the default/[set](https://github.com/Icinga/ipl-i18n/pull/20) ICU locale, or `Locale::getDefault()` which does the same.

It also makes sense to replace `strtotime()` calls with `new DateTime()` where the code will change.

Upon creating a PR, make sure to also reference (resolve) the following issues:

* https://github.com/Icinga/icingaweb2-module-director/issues/2828
* https://github.com/Icinga/icingaweb2-module-incubator/issues/11

raviks789:
fixes #2847

### DIFF
--- a/library/Director/Web/Table/ActivityLogTable.php
+++ b/library/Director/Web/Table/ActivityLogTable.php
@@ -2,15 +2,15 @@
 
 namespace Icinga\Module\Director\Web\Table;
 
-use gipfl\Format\LocalTimeFormat;
+use DateTime;
 use gipfl\IcingaWeb2\Link;
-use gipfl\IcingaWeb2\Table\ZfQueryBasedTable;
 use Icinga\Module\Director\Auth\Permission;
 use Icinga\Module\Director\Util;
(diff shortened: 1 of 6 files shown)


nilmerg (changes requested):

nilmerg:
Superseded by 428a49f4330c8863094e29582e106c3d8f9ef9f5

Issue #1203, opened 2017-09-25
editing multiple hosts removes template-choice imports


vita2:
i've just edited several hosts with the multiedit-form. after storing the changes my template-choice was removed from my host objects. it means the import line which imports the chosen template is gone.

## Your Environment
* Director version (System - About): commit f7353b66283ee3f11e15fd5d616a690ac7b72c4c
* Icinga Web 2 version and modules (System - About): 2.4.1
* Icinga 2 version (`icinga2 --version`): r2.7.1-1
* Operating System and version: Debian Jessie 8.9 (3.16.0-4-amd64)
* Webserver, PHP versions: nginx 1.6.2-5 + php5-fpm 5.6.30

vita2:
can you reproduce the behavior? unfortunately i hit this issue several times and had to correct the affected host-objects :disappointed_relieved:

PR #2867, opened 2024-02-08
Unhide imports from Choices when doing multiedit

arjanoosting:
When doing an multiedit the Choises are not shown but the corresponding Host templates are filtered from the imports.
And if you then do a save the imports are removed from the hosts. see issue #1203

This small patch unhides the filtered host templates when doing an multiedit.

The nicer solution is showing the choices as well in the multiedit, but that is much more complicated code change :)


bobapple:
@cla-bot check

lippserd:
@raviks789 Please check this PR.

PR #2891, opened 2024-07-17
ImportSourceRestApi: Fix strlen deprecation

### LINKED ISSUE #2889: Import Source: choosing "HTTP proxy" in Source Type "REST API" causes deprecation notice
## Expected Behavior
When setting the Proxy-Type in an import source of type "REST API" to "HTTP proxy" only the necessary fields "Proxy Address*", "Proxy Username" and "Proxy Password" should be visible.

## Current Behavior
After choosing "HTTP proxy" the following message appears (in addition to the proxy-fields):
Deprecated
: strlen(): Passing null to parameter #1 ($string) of type string is deprecated in
/usr/share/icingaweb2/modules/director/library/Director/Import/ImportSourceRestApi.php
on line
276

Go to
"Automation" -> Import source -> +Add -> Add a new source name, choose Source Type "REST API" -> choose "HTTP proxy"

## Your Environment

* Director version (System - About): 1.11.1
* Icinga Web 2 version and modules (System - About): 2.12.1
* Icinga 2 version (`icinga2 --version`): 2.14.2
* Operating System and version: SLES 15 SP5
* Webserver, PHP versions: Apache, 8.1.28

raviks789:
fix #2889

### DIFF
--- a/library/Director/Import/ImportSourceRestApi.php
+++ b/library/Director/Import/ImportSourceRestApi.php
@@ -273,11 +273,9 @@ class ImportSourceRestApi extends ImportSourceHook
                     ),
                 ]);
 
-                $passRequired = strlen($form->getSentOrObjectSetting('proxy_user')) > 0;
-
                 $form->addElement('storedPassword', 'proxy_pass', [
                     'label'    => $form->translate('Proxy Password'),
-                    'required' => $passRequired
+                    'required' => strlen((string) $form->getSentOrObjectSetting('proxy_user')) > 0
                 ]);
             }
         }


library/Director/Import/ImportSourceRestApi.php
```diff
@@ -273,11 +273,9 @@ protected static function addProxy(QuickForm $form)
                     ),
                 ]);
 
-                $passRequired = strlen($form->getSentOrObjectSetting('proxy_user')) > 0;
-
                 $form->addElement('storedPassword', 'proxy_pass', [
                     'label'    => $form->translate('Proxy Password'),
-                    'required' => $passRequired
+                    'required' => strlen((string) $form->getSentOrObjectSetting('proxy_user')) > 0
```
sukhwinder33445, line 278: I think you don't need `strlen`. `$form->getSentOrObjectSetting('proxy_user') ?: false` should be sufficient.

sukhwinder33445 (changes requested):

library/Director/Import/ImportSourceRestApi.php
raviks789: I don't think so, that is not a strict comparison.

sukhwinder33445: The returned value of the function can be a string, '' or null. And all these cases are covered in the suggested change? Or am I missing something?

raviks789: Did you test it? Try entering '0', for 'proxy_user field'.

sukhwinder33445: Ohh, I see, then you can use `??` instead of `?:`.

raviks789: As discussed offline this would not work as well.

PR #2916, opened 2024-09-13
DirectorDatalist: Match column types to join tables in postgres by explicit type casting

### LINKED ISSUE #2913: Deleting data list cause error when using PostgreSQL as backend
## Current Behavior
image

PostgreSQL is strict and does not do any magic typecasting. SQLServer does typecasting automagically. https://dev.mysql.com/doc/refman/8.4/en/type-conversion.html#:~:text=For%20example%2C%20MySQL%20automatically%20converts,as%20necessary%2C%20and%20vice%20versa.&text=It%20is%20also%20possible%20to,using%20the%20CAST()%20function. 

That is why the Problem does not appear by MySQL users.
## Possible Solution

change this

'l.id = dfs.setting_value',

to this

'CAST(l.id AS text) = dfs.setting_value',

in this file /usr/share/icingaweb2/modules/director/library/Director/Objects/DirectorDatalist.php (line 106):

create data list and use it. You should use PostgreSQL as backend. Try to delete the data list, you will get the same error above in the screenshot. 

## Your Environment

* Director version: 1.11.

raviks789:
fixes #2913

### DIFF
--- a/library/Director/Objects/DirectorDatalist.php
+++ b/library/Director/Objects/DirectorDatalist.php
@@ -94,6 +94,12 @@ class DirectorDatalist extends DbObject implements ExportInterface
 
         $db = $this->getDb();
 
+        if ($this->getConnection()->getDbType() === 'pgsql') {
+            $datalistJoin = 'l.id = CAST(dfs.setting_value AS INTEGER)';
+        } else {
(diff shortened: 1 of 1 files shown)


Wintermute2k6:
ref/NC/828101 is here any progress i can report ?

lippserd (approved):

PR #2923, opened 2024-09-30
added conversion from array to object for the directory to row property modifier

MisterMountain:
Without this fix modifying an import source fails, if you try to clone the row for every entry of your nested dictionary/hash structure due to a missing array to object conversion


lippserd (approved):

lippserd (changes requested):
@MisterMountain Please fix the [code sniffer complaints](https://github.com/Icinga/icingaweb2-module-director/actions/runs/11719251205/job/32642415893?pr=2923#step:7:12).

raviks789:
@MisterMountain, reproduced this issue. See the steps below:

  - Create an Import Source in Director (Source Type: SQL, any DB resource works).
  - Use this query, which returns one row with a text column holding some JSON:
  SELECT 'host1' AS `key`, '{"nginx": {"port": 80}, "logging": []}' AS services
  - On that Import Source, open the Modifiers tab and add two modifiers on the services column, in this order:
    a. "Decode a JSON string"
    b. "Clone the row for every entry of a nested Dictionary/Hash structure" (Key Property Name: service_name)
  - Open the Preview tab.
  - Result: the preview fails with Nested "logging" dictionary expected, got array (PropertyModifierDictionaryToRow.php:82), even though logging simply has no extra properties. The whole
    import source stops working, not just that one entry.

library/Director/PropertyModifier/PropertyModifierDictionaryToRow.php
```diff
@@ -78,6 +78,9 @@ public function transform($value)
         }
         $result = [];
         foreach ($value as $key => $properties) {
+            if (is_array($properties)) {
+            $properties = (object) $properties;
+            }
             if (! is_object($properties)) {
```
raviks789: The issue occurs due to empty array as empty array and empty object carry nothing in it they both can be treated the same, but the array with actual items in it should still fail.

```suggestion
            if (is_array($properties) && empty($properties)) {
                $properties = (object) $properties;
            }
            
if (! is_object($properties)) {
```

raviks789 (changes requested):

PR #2940, opened 2024-12-09
Fix service set member count while on director branch

raviks789:
While on director branch the service set member count is always displayed as 1 instead of actual count.

Before:
Screenshot 2024-12-09 at 18 21 25

After:
Screenshot 2024-12-09 at 18 23 53


lippserd:
Do you know why you've introduced the grouping in 3c7082ca in the first place?

raviks789:
> Do you know why you've introduced the grouping in [dcd3687](https://github.com/Icinga/icingaweb2-module-director/commit/dcd36877ea926be237ea38980524716f41315801) in the first place?

I don't exactly remember. But I think it was retained after I had introduced it to test the queries for this change.

PR #2941, opened 2024-12-10
IcingaHostForm: Fix crashing of form for invalid check command

### LINKED ISSUE #2854: Focusing out of the Check command field during the Host Object creation crashes site.
## Expected Behavior
I want to receive a error message under the field if I enter a wrong command instead of a crash where I have to redo everything.

## Current Behavior
If I focus out of the Check command field during the creation of a new object the whole site crashes.

* Try creating a host template
* give it a invalid value as check command
* Notice that the site completely crashes instead of giving you just an error message./

## Your Environment
* Director version (System - About): 1.11.0
* Icinga Web 2 version and modules (System - About): 2.10.5
* Icinga 2 version (`icinga2 --version`): r2.14.0-1
* Operating System and version: rhel 8.9
* Webserver, PHP versions: 7.4.33

raviks789:
fixes #2854

### DIFF
--- a/application/forms/IcingaHostForm.php
+++ b/application/forms/IcingaHostForm.php
@@ -2,6 +2,7 @@
 
 namespace Icinga\Module\Director\Forms;
 
+use Exception;
 use Icinga\Exception\AuthenticationException;
 use Icinga\Module\Director\Auth\Permission;
 use Icinga\Module\Director\Auth\Restriction;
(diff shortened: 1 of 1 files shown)


lippserd:
I'm not entirely convinced by the solution, as now the error is shown in both the check_command and has_agent fields, but I haven't found a better solution in a quick way, so let's just merge it.

PR #2943, opened 2024-12-11
Fix: Show services added for existing service sets in usage table while on director branch

raviks789:
Services added to the service sets while on director branch was previously not displayed in the template usage table.

Before:

https://github.com/user-attachments/assets/e1865fc7-30f6-4f3e-adc5-de9774bf051a

After:

https://github.com/user-attachments/assets/d656081d-6d8a-4deb-9ced-e19e5f713eae


raviks789:
@lippserd , I tested it in Postgres as well. It works.

Issue #2885, opened 2024-07-03
Missing unique key for constraint 'director_generated_config_activity' in the referenced table 'director_activity_log'


nilmerg:
Importing the MySQL schema on MySQL 8.4.0 shows this error:

> ERROR 6125 (HY000) at line 108: Failed to add the foreign key constraint. Missing unique key for constraint 'director_generated_config_activity' in the referenced table 'director_activity_log'

This is caused by a deprecation in v8.4.0: https://dev.mysql.com/doc/refman/8.4/en/mysql-nutshell.html#mysql-nutshell-deprecations (Nonstandard foreign keys)

mcsken:
Ran into this while trying to setup Icinga for the first time. Highly annoying as there is no way to find this info without analysing the SQL queries, and this is the only reference relating to Icinga and it won't show up on google searches.

I ended up fixing the SQL schema myself (with the help of AI since I am not highly experience in neither SQL or icinga):

> [!WARNING]  
> Working AI generated solution.

Replaced the following:
```sql
CREATE TABLE director_activity_log (
  id BIGINT(20) UNSIGNED AUTO_INCREMENT NOT NULL,
  object_type VARCHAR(64) NOT NULL,
  object_name VARCHAR(255) NOT NULL,
  action_name ENUM('create', 'delete', 'modify') NOT NULL,
  old_properties TEXT DEFAULT NULL COMMENT 'Property hash, JSON',
  new_properties TEXT DEFAULT NULL COMMENT 'Property hash, JSON',
  author VARCHAR(64) NOT NULL,
  change_time TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
  checksum VARBINARY(20) NOT NULL,
  parent_checksum VARBINARY(20) DEFAULT NULL,
  PRIMARY KEY (id),
  INDEX sort_idx (change_time),
  INDEX search_idx (object_name),
  INDEX search_idx2 (object_type(32), object_name(64), change_time),
  INDEX search_author (author),
  INDEX checksum (checksum),
  UNIQUE INDEX idx_checksum (checksum) -- Add this line
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
``` 

```sql
CREATE TABLE director_generated_config (
  checksum VARBINARY(20) NOT NULL COMMENT 'SHA1(last_activity_checksum;file_path=checksum;file_path=checksum;...)',
  director_version VARCHAR(64) DEFAULT NULL,
[...]

PR #2946, opened 2024-12-19
#2885 Fix MySQL8.4 nonstandard foreign keys deprecation

mcsken:
- See https://github.com/Icinga/icingaweb2-module-director/issues/2885 (and my comment on the issue).
 - Fix MySQL-8.4 nonstandard foreign keys deprecation.
 
> [!IMPORTANT]  
> I have no previous MySQL experience. This is working queries I generated with the help of AI, and tested locally that it works for my setup using the latest latest MySQL (version 8.4.3) while setting up the Director module on Icinga Web 2 for the first time. Someone with more experience in MySQL should review it, and ideally merge a fix ASAP because it leads to a broken setup for new users (and whoever upgrades MySQL to 8.4+) with a essentially useless SQL error message for a user, which I found no reference to when searching the documentation or online.

The main error the user would normally see looks something like:
```
Migration 0 failed (SQLSTATE[HY000]: General error: 6125 Failed to add the foreign key constraint. Missing unique key for constraint 'director_generated_config_activity' 'director_generated_config_activity' in the referenced table 'director_activity_log'...
```
director-error-messages

### DIFF
--- a/schema/mysql-migrations/upgrade_188.sql
+++ b/schema/mysql-migrations/upgrade_188.sql
@@ -1,0 +1,17 @@
+ALTER TABLE director_activity_log
+DROP INDEX checksum,
+ADD UNIQUE INDEX checksum (checksum);
+
+ALTER TABLE director_generated_config
+DROP FOREIGN KEY director_generated_config_activity;
+
+ALTER TABLE director_generated_config
+ADD CONSTRAINT director_generated_config_activity
+    FOREIGN KEY (last_activity_checksum)
+    REFERENCES director_activity_log (checksum)
+    ON DELETE RESTRICT
+    ON UPDATE RESTRICT;
+
+INSERT INTO director_schema_migration
+  (schema_version, migration_time)
+  VALUES (188, NOW());
--- a/schema/mysql.sql
+++ b/schema/mysql.sql
@@ -46,7 +46,7 @@ CREATE TABLE director_activity_log (
   INDEX search_idx (object_name),
   INDEX search_idx2 (object_type(32), object_name(64), change_time),
   INDEX search_author (author),
-  INDEX checksum (checksum)
+  UNIQUE INDEX checksum (checksum)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
 CREATE TABLE director_activity_log_remark (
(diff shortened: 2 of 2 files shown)


mcsken:
Not happy I had to sign something for an open source PR, but I signed it....

schema/mysql.sql
```diff
@@ -46,7 +46,8 @@ CREATE TABLE director_activity_log (
   INDEX search_idx (object_name),
   INDEX search_idx2 (object_type(32), object_name(64), change_time),
   INDEX search_author (author),
-  INDEX checksum (checksum)
+  INDEX checksum (checksum),
+  UNIQUE INDEX idx_checksum (checksum)
```
roskakori, line 50:
```suggestion
  UNIQUE INDEX idx_checksum (checksum)
```
Setting the index once should suffice.

bobapple:
@cla-bot check

mcsken:
I looked into this a bit more and tried to clean up the AI code:
 - Removed what I believe are duplicate/unnecessary indexes from the original commit.
 - Sticked to the projects original naming and formatting conventions.

Someone ofc should still review it properly, and potentially even test it because I only tested the original commit on my machine.

schema/mysql-migrations/upgrade_188.sql
```diff
@@ -0,0 +1,14 @@
+ALTER TABLE director_activity_log
+DROP INDEX checksum,
+ADD UNIQUE INDEX checksum (checksum);
+
+ALTER TABLE director_generated_config
```
lippserd, line 5: The constraint should be removed beforehand. Otherwise this will lead to duplicate key errors.

lippserd (changes requested):
@raviks789 Can you make the change and squash the commits so we can merge this as soon as possible?

schema/mysql-migrations/upgrade_188.sql
raviks789: Done

roskakori:
Maybe I am missing something, but what is the purpose of a foreign key using the `checksum` instead of the `id`? I'd imagine that the `checksum` is not guaranteed to be unique.

lippserd (approved):

raviks789:
@roskakori I cannot answer why the foreign key is a checksum, but after this change here the checksum would have to be unique. So the uniqueness is going to be guaranteed.

PR #2947, opened 2025-01-03
Fix fields shoud not be added if empty refs #2945

moreamazingnick:
This is properly not the best fix but since a host is basically the same objecttype as a hosttemplate and a host template has fields but a host has not this is the best I could come up with.

Best Regards
Nicolas


raviks789 (commented):
This is not a complete resolution. This will still not resolve updating of templates through cli commands or Rest API, if the exported data is being used to update them. As the exporter will still add the 'fields' property in these cases.

Issue #2347, opened 2021-06-30
Hidden datafields should forbid autocompletion


lazyfrosch:
## Expected Behavior

Our password fields are not meant to be autocompleted, so we should disable it.

## Current Behavior

Chrome tends to auto-fill passwords into datafields when they are hidden, when a single secret is stored for the Icinga site. Even when other data is pre-filled into the fields.

Firefox will usually ask, but not autofill it.

## Possible Solution

Set `autocompletion="off"` for the hidden inputs, maybe add this as setting for datafields in general?

## Your Environment

* Director version (System - About): 1.8.0
* Icinga Web 2 version and modules (System - About): 2.8.2

moreamazingnick:
autocomplete="off" gets ignored by Google Chrome

autocomplete = "new-password" works 

the changes can be made in 
> modules/director/application/views/helpers/FormStoredPassword.php

Pull request comming soon...

Best Regards
Nicolas

PR #2949, opened 2025-01-09
Disable autocomplete on hidden fields refs #2347

moreamazingnick:
It is a bad idea to let chrome store passwords but since some of us do that, there should not be any autocomplete on hidden fields in any director custom var field.

Best Regards 
Nicolas


raviks789:
> branch updated

Thank you

PR #2954, opened 2025-02-05
Fix job behavior during summer and winter time change

raviks789:
ref/IP/52977


schema/mysql.sql
```diff
@@ -2446,4 +2446,4 @@ CREATE TABLE branched_icinga_dependency (
 
 INSERT INTO director_schema_migration
   (schema_version, migration_time)
-  VALUES (187, NOW());
+  VALUES (189, NOW());
```
nilmerg, line 2449: Shouldn't 188 be next?

nilmerg (changes requested):
Do we know what will happen with the daemon if the database upgrade happens while it's active?

schema/mysql.sql
raviks789: I wanted to match it with the postgres migration version. And the postgres migration version was not updated to 188 previously even though there was a migration script for the version 188. So I thought it was right to use 189 instead of 188.

nilmerg: Okay. There was a gap between 78-82 anyway with MySQL.

schema/mysql-migrations/upgrade_189.sql
```diff
@@ -0,0 +1,17 @@
...
+
+ALTER TABLE director_job
+    DROP COLUMN ts_last_attempt,
+    DROP COLUMN ts_last_error,
+    CHANGE ts_last_attempt_tmp ts_last_attempt BIGINT(20) DEFAULT NULL,
+    CHANGE ts_last_error_tmp ts_last_error BIGINT(20) DEFAULT NULL;
+
+INSERT INTO director_schema_migration
+(schema_version, migration_time)
+VALUES (189, NOW());
```
lippserd, line 17: Newline at the end of the file is missing.

lippserd (changes requested):
Can you please also give a summary of the issue and how this PR fixes it.

schema/mysql.sql
lippserd: #2946 adds version 188.

lippserd (approved):

PR #2989, opened 2025-06-03
Remove locales

nilmerg:
I imported all of them on translate.icinga.com now. The file `translateMe` is obsolete since Icinga L10n.


raviks789 (approved):
LGTM!

PR #2994, opened 2025-06-27
Flatten nulls to empty string for preg_match in PropertyModifierRejectOrSelect

afoster:
In recent versions of PHP, `preg_match()` will throw a deprecation notice if null is passed as its `$subject` string parameter. This change  makes `matchesRegexp()` behave as it would have before when a null was implicitly casted, and is probably what admin users would expect when testing their patterns against a non-existant value versus an empty value.


afoster:
Hi, CLA signed as @afoster.

bobapple:
@cla-bot check

Issue #3007, opened 2025-09-03
Imported command is marked “Not in use” and variables are not shown in Overview


Saeid-Abadi:
### Expected behavior:
If Command A imports Command B (and uses B’s arguments/fields), B should not be shown as “Not in use”; Overview/Usage should reflect that Command B is effectively used through A, and variables inherited via import should appear as used.

### Current behavior: 
Command B displays “Not in use” and variables inherited from B are not shown as used/linked in Overview.

### Steps to reproduce:
* Create Command B with arguments/fields.
* Create Command A that imports B and uses its variables.
* Observe: Command B shows “Not in use” and the inherited variables/usage aren’t visible in Overview.

### Environment:
Icinga Director version: 1.11.5
 Icinga Web 2 version: 2.12.5
 
### Related: 
#2343 (imported command arguments/fields not shown).

PR #3009, opened 2025-09-19
Fix Command Inheritance not shown #3007

moreamazingnick:
shows commands and command templates and filters by commandname in the table views

Best Regards 
Nicolas


library/Director/Objects/IcingaCommand.php
```diff
@@ -183,8 +183,13 @@ public function countDirectUses()
             array('n' => 'icinga_notification'),
             array('cnt' => 'COUNT(*)')
         )->where('n.command_id = ?', $id);
+        $qc = $db->select()->from(
+            array('c' => 'icinga_command_inheritance'),
+            array('cnt' => 'COUNT(*)')
```
raviks789, line 188:
```suggestion
            [c' => 'icinga_command_inheritance'],
            ['cnt' => 'COUNT(*)']
```
We now use the short syntax of `array()` as it looks cleaner and modern

library/Director/Web/Controller/ObjectsController.php
```diff
@@ -474,6 +474,17 @@ protected function eventuallyFilterCommand(ZfQueryBasedTable $table)
...
+                    $table->getQuery()
+                        ->join(
+                            array('ici' => 'icinga_command_inheritance'),
+                            'ici.command_id = o.id',
+                            array()
+                        )->where(
+                            'ici.parent_command_id = ?',
+                            $command->getAutoincId()
+                        );
+                    break;
```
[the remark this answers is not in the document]
raviks789: Not a functional issue, this just makes it a bit easier to read

library/Director/Resolver/CommandUsage.php
```diff
@@ -89,14 +94,26 @@ protected function fetchFor($table, $rels, $objectTypes)
...
-        foreach ($rels as $rel) {
-            $query->orWhere("{$rel}_id = ?", $id);
+        if ($table === "command") {
+            $query = $this->db->select()->from(array('c' => 'icinga_' . $table), $columns)->join(array('ici' => 'icinga_command_inheritance'),
+                'ici.command_id = c.id',
+                array());
+
+            foreach ($rels as $rel) {
+                $query->orWhere("parent_{$rel}_id = ?", $id);
+            }
```
raviks789, line 110:
```suggestion
            $query = $this->db->select()->from(['c' => 'icinga_' . $table], $columns)->join(
                ['ici' => 'icinga_command_inheritance'],
                'ici.command_id = c.id',
                []
            )->where('ici.parent_command_id = ?', $id);
```

IMO, you could rewrite the `where` as suggested, just makes it a bit easier to understand the code

library/Director/Resolver/CommandUsage.php
```diff
@@ -89,14 +94,26 @@ protected function fetchFor($table, $rels, $objectTypes)
     {
         $id = $this->command->getAutoincId();
 
+        //return (object) array('object'=>1);
```
raviks789, line 97: This comment seems unnecessary here, remove it.

raviks789 (changes requested):

raviks789 (changes requested):

library/Director/Resolver/CommandUsage.php
moreamazingnick: this would render the $rels var useless, and might break something else

raviks789: The `$rels` variable comes from the `$map` variable. You could give an empty array there, as the relation in the map is basically used to build the column `{rel}_id` to filter the objects. And since for `command` you filter  its inheritance table. You could directly filter it in the query.

Or you could do the below, where you add the join and the filter to the query only if the table is 'command', and everything remains same. This solution also results in minimal changes to the existing code. You should still give empty array for command (`'command' => []`) in `$map` variable. Also, looking at it again I find this solution better.

```php
        $query = $this->db->select()->from(['c' => "icinga_$table"], $columns);
        if ($table === "command") {
            $query->join(
                ['ici' => 'icinga_command_inheritance'],
                'ici.command_id = c.id',
                []
            )->where('ici.parent_command_id = ?', $id);
        }

        foreach ($rels as $rel) {
            $query->orWhere("{$rel}_id = ?", $id);
        }
```

library/Director/Resolver/CommandUsage.php
```diff
@@ -44,6 +44,7 @@ public function getLinks()
             'host'         => ['check_command', 'event_command'],
             'service'      => ['check_command', 'event_command'],
             'notification' => ['command'],
+            'command'      => ['command'],
```
raviks789, line 47:
```suggestion
            'command'      => [],
```

This could be an empty array, as this is basically used to build the command id column for the table to filter the objects using the corresponding command.

raviks789 (changes requested):

library/Director/Resolver/CommandUsage.php
moreamazingnick: I think I added that because a command can inherit from a command but after so much time I am not sure anymore

raviks789:
@moreamazingnick , could you provide me with [Allow edits by maintainers](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests)? So that I could make the suggested edits, test and subsequently merge the request.

moreamazingnick:
I can't because the checkbox is missing.
Bildschirmfoto 2026-09-14 um 15 09 24

I tried to enable   "maintainer_can_modify": true using gh cli

raviks789:
Closing this PR as opened a new PR for this #3113

PR #3024, opened 2026-01-12
Don't require permissions for custom var rendering hooks

lippserd:
Previously, at least general access permissions for the module were required in
order to execute the custom variable rendering hooks. This has been relaxed so
that users who do not have permissions for the Director module at all can now
also benefit from the custom variable rendering hooks, e.g., that Director
labels instead of custom variable names from configuration are displayed.


lippserd:
ref/IP/59392

PR #3043, opened 2026-03-02
Adapt to `react/promise` v3 and `react/event-loop` v1.6

jrauh01:
### Summary

- Upgrade `react/promise` to **v3**:
  - replace removed `ExtendedPromiseInterface` with `PromiseInterface`
  - pass required arguments to `resolve()` and `reject()`
  - rename `otherwise()`/`always()` to `catch()`/`finally()`
  - remove dead `FinishedProcessState` catch handler that can never trigger in v3 (because it doesn't implement the `Throwable` interface); `gipfl\IcingaCliDaemon\IcingaCli` already rejects with an Exception wrapping the process state reason
- Upgrade `react/event-loop` to **v1.6**:
  - replace deprecated `Factory::create()` with `Loop::get()`

### References

https://github.com/reactphp/promise/releases/tag/v3.0.0
https://github.com/reactphp/event-loop/releases/tag/v1.2.0

requires: 

- [x] Icinga/icinga-php-thirdparty#93

### DIFF
--- a/library/Director/Daemon/DbBasedComponent.php
+++ b/library/Director/Daemon/DbBasedComponent.php
@@ -3,17 +3,19 @@
 namespace Icinga\Module\Director\Daemon;
 
 use Icinga\Module\Director\Db;
+use React\Promise\PromiseInterface;
 
 interface DbBasedComponent
 {
     /**
      * @param Db $db
-     * @return \React\Promise\ExtendedPromiseInterface;
+     *
+     * @return PromiseInterface;
      */
     public function initDb(Db $db);
 
     /**
-     * @return \React\Promise\ExtendedPromiseInterface;
+     * @return PromiseInterface;
      */
     public function stopDb();
 }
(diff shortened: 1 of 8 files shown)


lippserd (approved):

library/Director/Daemon/DbBasedComponent.php
```diff
@@ -3,17 +3,19 @@
...
 use Icinga\Module\Director\Db;
+use React\Promise\PromiseInterface;
 
 interface DbBasedComponent
 {
     /**
      * @param Db $db
-     * @return \React\Promise\ExtendedPromiseInterface;
+     *
+     * @return PromiseInterface;
```
raviks789: @jrauh01 could you remove the semicolon from the return statement in the PHPDoc

jrauh01: Done

raviks789 (approved):

PR #3046, opened 2026-03-04
Add `--kickstart` flag to `icingacli director daemon run`

jrauh01:
Allow running migrations, kickstart and config deployment as part of the daemon startup by passing `--kickstart`.

This simplifies container entrypoints and systemd units by consolidating the typical startup sequence into a single command. When kickstart is not configured, the process aborts with an error to signal a misconfiguration. If kickstart
  is configured but not required (already done before), that step is simply skipped and the daemon starts normally.

Calling
```
icingacli director daemon run --kickstart
```
 is similar to
```
icingacli director migration run \
&& icingacli director kickstart required \
&& icingacli director kickstart run \
&& icingacli director config deploy \
&& icingacli director daemon run
```

A kickstart run can delete Endpoint, Zone or Command objects that were imported before but are no longer present on the Icinga 2 master. Since `--kickstart` is meant to run unattended on every startup, it refuses to run if the Director
  DB already has such objects, and exits with an error instead.

### DIFF
--- a/application/clicommands/DaemonCommand.php
+++ b/application/clicommands/DaemonCommand.php
@@ -4,6 +4,11 @@ namespace Icinga\Module\Director\Clicommands;
 
 use Icinga\Module\Director\Cli\Command;
 use Icinga\Module\Director\Daemon\BackgroundDaemon;
+use Icinga\Module\Director\Db;
+use Icinga\Module\Director\Db\Migrations;
+use Icinga\Module\Director\Deployment\ConditionalDeployment;
+use Icinga\Module\Director\IcingaConfig\IcingaConfig;
+use Icinga\Module\Director\KickstartHelper;
 
 class DaemonCommand extends Command
 {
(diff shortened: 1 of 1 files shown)


application/clicommands/DaemonCommand.php
```diff
@@ -12,15 +17,63 @@ class DaemonCommand extends Command
...
+     *                config before starting the daemon
      */
     public function runAction()
     {
         $this->app->getModuleManager()->loadEnabledModules();
+        $dbResource = $this->params->get('db-resource');
+        if ($this->params->get('kickstart')) {
+            $this->runKickstart($dbResource);
+        }
         $daemon = new BackgroundDaemon();
```
raviks789, line 34: introduce a new line after if block.

```suggestion
        }

$daemon = new BackgroundDaemon();
```

application/clicommands/DaemonCommand.php
```diff
@@ -12,15 +17,63 @@ class DaemonCommand extends Command
...
+        // Like icingacli director migration run
+        (new Migrations($db))->applyPendingMigrations();
+
+        // Like icingacli director kickstart required
+        $kickstart = new KickstartHelper($db);
+        if (! $kickstart->isConfigured()) {
+            echo "Kickstart has not been configured\n";
+            exit(1);
+        }
+        if (! $kickstart->isRequired()) {
```
raviks789, line 54: same here

application/clicommands/DaemonCommand.php
```diff
@@ -12,15 +17,63 @@ class DaemonCommand extends Command
...
+            echo "Kickstart configured, execution is not required\n";
+            exit(1);
+        }
+        if ($this->isVerbose) {
```
raviks789, line 58: same here

application/clicommands/DaemonCommand.php
```diff
@@ -12,15 +17,63 @@ class DaemonCommand extends Command
...
+        $dbResource = $this->params->get('db-resource');
+        if ($this->params->get('kickstart')) {
+            $this->runKickstart($dbResource);
+        }
         $daemon = new BackgroundDaemon();
-        if ($dbResource = $this->params->get('db-resource')) {
+        if ($dbResource) {
             $daemon->setDbResourceName($dbResource);
         }
         $daemon->run();
```
raviks789, line 38: same here

application/clicommands/DaemonCommand.php
```diff
@@ -12,15 +17,63 @@ class DaemonCommand extends Command
...
     }
+
+    protected function runKickstart(?string $dbResource)
+    {
+        $db = $dbResource === null ? $this->db() : Db::fromResourceName($dbResource);
+
```
raviks789: You can set the db property here itself

```suggestion
        $this->db = $db;
```

application/clicommands/DaemonCommand.php
```diff
@@ -12,15 +17,63 @@ class DaemonCommand extends Command
...
+        }
+
+        // Like icingacli director kickstart run
+        $this->raiseLimits();
+        $kickstart->loadConfigFromFile()->run();
+
+        // Like icingacli director config deploy
+        $config = IcingaConfig::generate($db);
+        $checksum = $config->getHexChecksum();
+        $deployer = new ConditionalDeployment($db, $this->api());
```
raviks789, line 69: Since $this->api() calls $this->db() and fetches Icinga Endpoint for the default database instead of the passed --db-resource, there might be a chance that the config would be pushed to the wrong Icinga Endpoint. Make sure the config and the endpoint it's pushed to always come from the same resource.

```suggestion
        $deployer = new ConditionalDeployment($db, $db->getDeploymentEndpoint()->api());
```

raviks789 (changes requested):

application/clicommands/DaemonCommand.php
jrauh01: Done

application/clicommands/DaemonCommand.php
jrauh01: Done

application/clicommands/DaemonCommand.php
jrauh01: Done

raviks789 (approved):

lippserd (approved):

PR #3062, opened 2026-04-01
Changelog for Release version 1.11.8

raviks789:


module.info
```diff
@@ -1,6 +1,8 @@
 Name: Icinga Director
-Version: 1.11.7
-Depends: reactbundle (>=0.9.0), ipl (>=0.5.0), incubator (>=0.22.0)
+Version: 1.11.8
+Requires:
+  Libraries: icinga-php-library (>=0.5.0), icinga-php-thirdparty (>=0.12.1)
```
nilmerg: There's no v0.5.0 of icinga-php-library. Since PHP 7 support is still required for 1.11, require v0.14.3 instead.

nilmerg (changes requested):

module.info
nilmerg: Sorry, v0.14.**2**

nilmerg (approved):

PR #3075, opened 2026-05-18
Kickstart with mock data

raviks789:
Requires #3046

### DIFF
--- a/application/clicommands/DaemonCommand.php
+++ b/application/clicommands/DaemonCommand.php
@@ -4,6 +4,11 @@ namespace Icinga\Module\Director\Clicommands;
 
 use Icinga\Module\Director\Cli\Command;
 use Icinga\Module\Director\Daemon\BackgroundDaemon;
+use Icinga\Module\Director\Db;
+use Icinga\Module\Director\Db\Migrations;
+use Icinga\Module\Director\Deployment\ConditionalDeployment;
+use Icinga\Module\Director\IcingaConfig\IcingaConfig;
+use Icinga\Module\Director\KickstartHelper;
 
 class DaemonCommand extends Command
 {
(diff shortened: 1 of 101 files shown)


application/clicommands/DaemonCommand.php
```diff
@@ -90,24 +156,129 @@ protected function runKickstart(?string $dbResource): void
...
 
+        foreach ($sources as $source) {
+            if ($source->runImport()) {
+                echo "New data has been imported\n";
+            } elseif ($source->get('import_state') === 'failing') {
+                $this->fail(
+                    "Import '%s' failed: %s",
+                    $source->get('source_name'),
+                    $source->get('last_error_message')
+                );
```
lippserd: Check `import_state` before treating the return value as success.
`ImportSource::checkForChanges()` sets `$hadChanges = true` before
persistence and can still return `true` after catching a storage error
and setting `import_state` to `failing`. This skips the failure branch
here and lets startup continue into sync and deployment with stale data.

Add a regression test for a persistence failure after changes are
detected, asserting that neither sync nor deployment starts.

```suggestion
            $hasChanges = $source->runImport();
            if ($source->get('import_state') === 'failing') {
                $this->fail(
                    "Import '%s' failed: %s",
                    $source->get('source_name'),
                    $source->get('last_error_message')
                );
            } elseif ($hasChanges) {
                echo "New data has been imported\n";
```

doc/75-Background-Daemon.md
```diff
@@ -42,31 +42,44 @@ You now can start the Background daemon like any other service on your Linux sys
...
+kickstart already ran, that step gets skipped and the daemon starts as
+normal. If kickstart was never set up at all, the command stops with an
+error instead of starting, since that means the install isn't ready yet.
+`--import` and `--run-sync` don't need kickstart to be configured at
+all, so they also work on a setup seeded purely from a basket snapshot
+with no Icinga 2 API to kickstart from. `--deploy` doesn't need
+kickstart either, but it does need an Endpoint with an API user already
+in the Director DB to deploy to, whether that came from kickstart, a
+basket restore or a sync run. Without one, it fails with a clear error
+instead of deploying.
```
lippserd: Basket snapshots do not contain Endpoints or API users, so restoring a
basket cannot supply the deployment endpoint and credentials described
here. Document that these must be provisioned separately or through
kickstart.

```suggestion
in the Director DB. Provision these separately or run kickstart, since
basket snapshots do not contain Endpoints or API users. Without a
deployment endpoint, deployment fails with a clear error.
```

lippserd (changes requested):

lippserd (approved):

Issue #932, opened 2017-05-10
being able to edit “display name” in the “assign Service Section” … and being able to assign Services to Service Sets out of a apply rule instead of using the “assign where” section.


rai-n-air:
Example: I have to check a mysql server with several instances:
srv1, srv2, srv3 
The information about these Instances (Ports…etc) are stored to Custom Variables at the host, imported by “Decode a JSON string” like this:
{
"MYSQLPORT":["0815", "0816", "0817", "0818"],
"MYSQLINST":["srv1", "srv2", "srv21", "srv2"]
}
Out of this information I am able to generate several checks (Port, Filesystem, Connection…)  for each entry like this:
check_fs /bla/srv1
check_fs /bla/srv2
check_fs /bla/srv3
check_fs /bla/srv4
check_connection srv1
check_connection srv2
check_connection srv3
check_connection srv4
…

I only use the checks “check_fs” and “check_connection”.
To achive this I do the following:
Create a check
Create a assign rule like this:
In the GUI:
Name: check_fs
Apply for: MYSQLINST(MYSQLINST)
Assign where " MYSQLINST != ""

The resulting config:
Config preview: check_fs
Show resolved
zones.d/director-global/service_apply.conf
apply Service "check_fs" for (config in host.vars.MYSQLINST) {
    import "check_fs"

    assign where host.vars.MYSQLINST != ""

    import DirectorOverrideTemplate
}

This results in the following entry in the Director DB:

MariaDB [director]> select object_name,object_type,display_name,service_set_id,apply_for,assign_filter from icinga_service where object_name = 'check_fs';
+-------------+-------------+--------------+----------------+---------------------+-----------------------------+
| object_name | object_type | display_name | service_set_id | apply_for           | assign_filter               |
+-------------+-------------+--------------+----------------+---------------------+-----------------------------+
| check_fs    | template    | NULL         |           NULL | NULL                | NULL                        |
| check_fs    | apply       | NULL         |           NULL | host.vars.MYSQLINST | host.vars.MYSQLINST!=%22%22 |
+-------------+-------------+--------------+----------------+---------------------+-----------------------------+
2 rows in set (0.00 sec)

If I now change the display name from “NULL” to “check_fs $config$” it shows the iteration in the checkname in Icingaweb2. Nice (thank you Thilo Wening@Netways)!
It would be great if this is possible in the Director GUI instead of injecting the Director DB.

[...]

PR #3078, opened 2026-06-17
Expose `display_name` field in the service form

### LINKED ISSUE #608: [dev.icinga.com #13233] Missing "display_name" field in Service tab

### LINKED ISSUE #2659: Add "display name" for service apply rules in the director UI
Please add the `display_name` field for the service apply rules to the Director UI.

The `display_name` is already available when using the director API and works as expected. It is just missing in the UI.

Positive side effect: Currently service apply rules do not have unique names. The point being - the user might want to have same service name with different rules. From my perspective adding `display_name` would also allow to make the `object_name` for the service apply rules **unique**, as it is the case for the other object types. This would solve many problems for tools which are using the director API, trying to work with the `id` that is stored in the director. But this breaks decoupled nature of such tools, as they need to store the dynamically generated `id` information.

You absolutely do not need to make the `object_name` unique now, but the user might choose to do so if he likes. And I believe most of the users would like to maintain the service apply rules that way.

### LINKED ISSUE #2957: cannot update/modify the display_name via web-form
Hello team, I'd like to ask to add additional field into the web-form to be able to edit/modify/update the display_name once it is set to different string as the Host's/Service's name. 
Currently this is possible only via import/sync rule or using the REST API what is not comfortable for the regular users.
Thanks.

Fantu:
The `display_name` property already exists on `IcingaService` and is rendered into the generated config, but the web form never exposed it, so it could only be set through manual SQL or the REST API. This adds a `Display name` text field to the service form, mirroring the existing host form behaviour.

A dedicated `addDisplayNameElement()` is added and wired into all three setup paths (service objects and apply rules, host-related services and service-set services), skipping templates just like the host form does.

Fixes #608
Fixes #2659
Fixes #2957
Refs #932
Refs #2509

### DIFF
--- a/application/forms/IcingaServiceForm.php
+++ b/application/forms/IcingaServiceForm.php
@@ -424,6 +424,7 @@ class IcingaServiceForm extends DirectorObjectForm
         $forceCommandElements = $this->hasPermission(Permission::ADMIN);
 
         $this->addNameElement()
+             ->addDisplayNameElement()
              ->addHostObjectElement()
              ->addImportsElement()
              ->addChoices('service')
@@ -570,6 +571,7 @@ class IcingaServiceForm extends DirectorObjectForm
         }
 
         $this->addNameElement()
+             ->addDisplayNameElement()
              ->addChoices('service')
              ->addDisabledElement()
              ->addGroupsElement()
@@ -612,6 +614,7 @@ class IcingaServiceForm extends DirectorObjectForm
         }
 
         $this->addNameElement()
+             ->addDisplayNameElement()
              ->addDisabledElement()
              ->addGroupsElement()
              ->groupMainProperties();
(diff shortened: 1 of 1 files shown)


Fantu:
cla was signed

bobapple:
@cla-bot check

Fantu:
@bobapple thanks

raviks789:
@Fantu, please rebase the branch with master.

Fantu:
@raviks789 thanks for the reply. I did the rebase, was clean and from a fast look with this small patch there shouldn't be any issues with other modifications (I even applied clean to an old version running on a production server without problems), but I can run a more thorough check if needed.

raviks789 (approved):

PR #3090, opened 2026-07-09
IcingaServiceSet: Fix duplicate services when importing a set with a changed member UUID

### LINKED ISSUE #3089: Duplicate services in a Service Set after a Basket import with a different UUID
## Describe the bug

Restoring a Basket that contains a Service Set whose member service carries a
different UUID than the one already stored creates a **duplicate** service inside
the set. Afterwards two `icinga_service` rows share the same `object_name` and
`service_set_id` and differ only in their UUID.

Two things combine to cause this:

* `IcingaServiceSet::setServices()` matches existing members **by UUID only**,
  with no fallback to the object name. A member whose UUID changed is therefore
  not recognised, and a new service is created.
* `IcingaServiceSet::storeRelatedServices()` is meant to delete the now orphaned
  old member. Its delete loop enumerates the set members through
  `fetchServices()` -> `ServiceSetQueryBuilder::fetchServicesWithQuery()`, which
  uses the `object_name` as its key. Therefore, two same-named rows collapse into one
  entry, so the orphan is never seen by the delete loop and survives.

The database does not prevent it either: the only relevant unique key is
`icinga_service (object_name, host_id)`, and for Service Set members `host_id` is
`NULL`. Because NULLs are treated as distinct, two members with the same name in
the same set do not collide.

The duplicate **is visible** in the Service Set's service list in the GUI, which
renders the members through the raw, non-deduplicated query
(`IcingaServiceSetServiceTable`). It is **hidden** in the config preview, in the
deployed configuration and in a Basket dump, because those go through
`fetchServices()` and the name-deduplicating query, which shows the member only
once. Which of the two copies wins that deduplication (and is therefore rendered)
is not deterministic (the query has no `ORDER BY`), so when the two copies differ,
Director can silently deploy the stale one.

## To Reproduce

Two minimal Baskets that describe the **same** Service Set (same UUID) with
the **same** member name, differing only in the member's UUID.

`member1.json`:
```json
{
    "ServiceSet": {
        "___TEST___dupset": {
            "assign_filter": "host.name=%22*%22",
            "description": "Reproducer for duplicate Service Set members",
            "object_name": "___TEST___dupset",
            "object_type": "template",
            "services": [
                {
                    "object_name": "___TEST___dupsvc",
                    "object_type": "object",
                    "uuid": "aaaaaaaa-1111-4111-8111-111111111111",
                    "vars": { "___TEST___marker": "from-basket-1" }
                }
            ],
            "uuid": "5e750000-0000-4000-8000-000000000001"
        }
    }
}
```

`member2.json` (identical apart from the member `uuid` and the marker value):
```json
{
    "ServiceSet": {
        "___TEST___dupset": {
            "assign_filter": "host.name=%22*%22",
            "description": "Reproducer for duplicate Service Set members",
            "object_name": "___TEST___dupset",
            "object_type": "template",
            "services": [
                {
                    "object_name": "___TEST___dupsvc",
                    "object_type": "object",
                    "uuid": "bbbbbbbb-2222-4222-8222-222222222222",
                    "vars": { "___TEST___marker": "from-basket-2" }
                }
            ],
            "uuid": "5e750000-0000-4000-8000-000000000001"
        }
    }
}
```

1. Restore the first Basket:
   `icingacli director basket restore < member1.json`
2. Confirm the set has exactly one member:
   ```sql
   SELECT s.id, LOWER(HEX(s.uuid)) AS uuid, s.object_name
   FROM icinga_service s
   JOIN icinga_service_set ss ON s.service_set_id = ss.id
   WHERE ss.object_name = '___TEST___dupset' ORDER BY s.id;
   ```
   Should return one row, for example:
    ```
    +------+----------------------------------+------------------+-------------+
    | id   | uuid                             | object_name      | object_type |
    +------+----------------------------------+------------------+-------------+
    | 1517 | aaaaaaaa111141118111111111111111 | ___TEST___dupsvc | object      |
    +------+----------------------------------+------------------+-------------+
    ```
3. Restore the second Basket (same member name, new member UUID):
   `icingacli director basket restore < member2.json`
4. Run the SQL query again, now it should return 2 rows. For example:
    ```
    +------+----------------------------------+------------------+-------------+
    | id   | uuid                             | object_name      | object_type |
    +------+----------------------------------+------------------+-------------+
    | 1517 | aaaaaaaa111141118111111111111111 | ___TEST___dupsvc | object      |
    | 1518 | bbbbbbbb222242228222222222222222 | ___TEST___dupsvc | object      |
    +------+----------------------------------+------------------+-------------+
    ```
    And the Service Set's service list in the GUI now shows the service twice.

## Expected behavior

After the second restore the Service Set contains `___TEST___dupsvc` exactly once,
reflecting the member from the Basket (UUID `bbbb...` with its properties). The
previously stored member (UUID `aaaa...`) no longer exists. This is the same
outcome as importing a Basket whose member UUID was unchanged: the Basket's member
wins and no duplicate `icinga_service` rows remain.

## Screenshots

The Service Set's service list in the GUI shows `___TEST___dupsvc` twice after the
second restore:
Image

While the config preview shows it once:
Image

## Your Environment

* Director version (System - About): v1.11.9
* Icinga Web 2 version and modules (System - About): 2.14.0
* Icinga 2 version (`icinga2 --version`): r2.16.3-1
* Webserver, PHP versions: Apache/2.4.37, PHP 8.4.23

## Additional context

Root-cause pointers for a fix:

* `library/Director/Objects/IcingaServiceSet.php`, `setServices()` matches members
  by UUID only.
* `library/Director/Objects/IcingaServiceSet.php`, `storeRelatedServices()` deletes
  orphans by iterating `fetchServices()`.
* `library/Director/Data/Db/ServiceSetQueryBuilder.php`,
  `fetchServicesWithQuery()` keys its result by `object_name`, which collapses two
  same-named rows and hides the orphan from the delete loop.

Suggested fix: have the delete loop in `storeRelatedServices()` enumerate the raw,
non-deduplicated member rows (for example `SELECT uuid FROM icinga_service WHERE
service_set_id = ?`) so every orphaned member UUID is visible for deletion. The
member from the Basket then wins and the previously stored member is removed, the
same outcome as a matching-UUID import.

NavidSassan:
Fixes https://github.com/Icinga/icingaweb2-module-director/issues/3089

### DIFF
--- a/library/Director/Data/Db/ServiceSetQueryBuilder.php
+++ b/library/Director/Data/Db/ServiceSetQueryBuilder.php
@@ -119,6 +119,27 @@ class ServiceSetQueryBuilder
         return $services;
     }
 
+    /**
+     * Fetch the raw UUIDs (binary, 16 bytes) of all services matched by the query,
+     * one entry per row.
+     *
+     * Unlike fetchServicesWithQuery(), this does NOT deduplicate by object_name, so
+     * callers that need to see every stored row (for example to reconcile and delete
+     * orphaned Service Set members) do not lose same-named duplicates.
+     *
+     * @return string[]
+     */
+    public function fetchServiceUuids(\Zend_Db_Select $query): array
+    {
+        static::resetQueryProperties($query);
+        $uuids = [];
+        foreach ($this->connection->getDbAdapter()->fetchCol($query) as $uuid) {
+            $uuids[] = DbUtil::binaryResult($uuid);
+        }
+
+        return $uuids;
+    }
+
     protected function getColumns()
     {
         return [
--- a/library/Director/Objects/IcingaServiceSet.php
+++ b/library/Director/Objects/IcingaServiceSet.php
@@ -166,9 +166,34 @@ class IcingaServiceSet extends IcingaObject implements ExportInterface
             $service->store();
         }
 
-        foreach ($this->fetchServices() as $service) {
-            if (!isset($seen[$service->getUniqueId()->getBytes()])) {
-                $service->delete();
+        // Delete members that are no longer part of this set. Two deliberate choices:
+        //
+        // 1. Enumerate the raw member UUIDs (one entry per row) via fetchServiceUuids()
+        //    instead of fetchServices(), which keys its result by object_name. Two
+        //    same-named rows collapse into a single entry there, hiding a duplicated
+        //    or orphaned member from this loop and leaving it behind.
+        //
+        // 2. Build the query with an explicit null branch UUID (no branch view).
+        //    Baskets, the code path that triggers this reconcile, always restore
+        //    straight into the main Director DB: BasketSnapshot::restoreObjects()
+        //    calls $object->store() on the plain connection, never through the
+        //    branch-aware DbObjectStore. So there is no branch to honour here, and the
+        //    members we reconcile live in the main icinga_service table. That also
+        //    matches the delete itself: IcingaService::delete() is a physical delete
+        //    against icinga_service, so the read must target the same rows as the
+        //    write. (A branch view would additionally list branch-only rows that are
+        //    absent from icinga_service and hide rows the branch marks deleted.)
+        //    Conversely, a Service Set edited inside a branch goes through
+        //    DbObjectStore::store(), which returns early and never reaches
+        //    store() -> onStore() -> here, so this method only ever runs against the
+        //    main DB anyway.
+        $builder = new ServiceSetQueryBuilder($this->getConnection(), null);
+        foreach ($builder->fetchServiceUuids($builder->selectServicesForSet($this)) as $uuid) {
+            if (! isset($seen[$uuid])) {
+                IcingaService::loadWithUniqueId(
+                    Uuid::fromBytes($uuid),
+                    $this->getConnection()
+                )->delete();
             }
         }
     }
--- a/test/php/library/Director/Objects/IcingaServiceSetTest.php
+++ b/test/php/library/Director/Objects/IcingaServiceSetTest.php
@@ -2,8 +2,10 @@
 
 namespace Tests\Icinga\Module\Director\Objects;
 
+use Icinga\Module\Director\Objects\IcingaService;
 use Icinga\Module\Director\Objects\IcingaServiceSet;
 use Icinga\Module\Director\Test\IcingaObjectTestCase;
+use Ramsey\Uuid\Uuid;
 
 class IcingaServiceSetTest extends IcingaObjectTestCase
 {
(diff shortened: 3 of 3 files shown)


library/Director/Objects/IcingaServiceSet.php
```diff
@@ -166,9 +166,34 @@ protected function storeRelatedServices()
...
+        //    branch-aware DbObjectStore. So there is no branch to honour here, and the
+        //    members we reconcile live in the main icinga_service table. That also
+        //    matches the delete itself: IcingaService::delete() is a physical delete
+        //    against icinga_service, so the read must target the same rows as the
+        //    write. (A branch view would additionally list branch-only rows that are
+        //    absent from icinga_service and hide rows the branch marks deleted.)
+        //    Conversely, a Service Set edited inside a branch goes through
+        //    DbObjectStore::store(), which returns early and never reaches
+        //    store() -> onStore() -> here, so this method only ever runs against the
+        //    main DB anyway.
```
raviks789, line 189: Shorten this comment to two or three lines maximum, it is way too long for this change.

library/Director/Data/Db/ServiceSetQueryBuilder.php
```diff
@@ -119,6 +119,27 @@ public function fetchServicesWithQuery(\Zend_Db_Select $query)
         return $services;
     }
 
+    /**
+     * Fetch the raw UUIDs (binary, 16 bytes) of all services matched by the query,
+     * one entry per row.
+     *
+     * Unlike fetchServicesWithQuery(), this does NOT deduplicate by object_name, so
+     * callers that need to see every stored row (for example to reconcile and delete
+     * orphaned Service Set members) do not lose same-named duplicates.
```
raviks789, line 128: A onenline note that it doesnot merge same-named rows should be sufficient.

```suggestion
     * Unlike fetchServicesWithQuery(), doesn't merge same-named rows
```

test/php/library/Director/Objects/IcingaServiceSetTest.php
```diff
@@ -107,6 +109,114 @@ public function testAddingServicesToSet()
...
+     * Regression test: importing a Set member with a *new* UUID (as a regenerated
+     * Basket does) must not leave a duplicate behind.
+     *
+     * IcingaServiceSet::setServices() matches members by UUID only, so a changed UUID
+     * creates a new service. storeRelatedServices() is expected to delete the now
+     * orphaned old member, but it enumerates via the name-deduplicating fetchServices()
+     * (ServiceSetQueryBuilder::fetchServicesWithQuery(), which keys results by
+     * object_name). Two same-named rows collapse into one, hiding the orphan from the
+     * deletion loop, so it survives and the member ends up duplicated.
+     */
```
raviks789, line 161: No need to over explain everything again, the one liner below must be sufficient.

```suggestion
    /**
     * Regression: A member re-imported under a new uuid must replace, not duplicate, the old one.
     */
```

library/Director/Data/Db/ServiceSetQueryBuilder.php
```diff
@@ -119,6 +119,27 @@ public function fetchServicesWithQuery(\Zend_Db_Select $query)
...
+     *
+     * @return string[]
```
raviks789, line 130:
```suggestion
     * orphaned Service Set members) do not lose same-named duplicates.
     * 
     * @param Zend_Db_Select $query
     * 
     * @return string[]
```

library/Director/Data/Db/ServiceSetQueryBuilder.php
```diff
@@ -119,6 +119,27 @@ public function fetchServicesWithQuery(\Zend_Db_Select $query)
...
+     */
+    public function fetchServiceUuids(\Zend_Db_Select $query): array
```
raviks789, line 132: Add `use Zend_Db_Select;` to the use block and change the signature to `public function fetchServiceUuids(Zend_Db_Select $query): array`

raviks789 (changes requested):

PR #3097, opened 2026-07-31
IcingaTemplateChoiceForm: Fix associated service template lookup

### LINKED ISSUE #2839: icinga_service has a multicolumn key, array required (DbObject.php:1024) when saving choice template
## Expected Behavior
It should be possible to save a choice template without an error.

## Current Behavior
When saving a new choice or an existing one the following error is shown: 
`icinga_service has a multicolumn key, array required (DbObject.php:1024)`

Screenshot of an existing and working choice template after klicking the `Store` button.
grafik

## Possible Solution
None

Create a choice template and ty save it. 

## Your Environment

* Director version (System - About): 1.11.0
* Icinga Web 2 version and modules (System - About): 2.12.0
* Icinga 2 version (`icinga2 --version`): 2.14.0
* Operating System and version: Ubuntu 20.04
* Webserver, PHP versions: php 7.4.3-4ubuntu2.19

DanielVdWP:
Service template choices currently submit the associated template by object name. This causes the generic relation resolver to call IcingaService::load() with a scalar value, while IcingaService uses a multi-column key, resulting in: icinga_service has a multicolumn key, array required.

Use the existing required_template_id foreign key directly for the Associated Template form element instead. Available Choices continue to use object names as before.

A regression test verifies that service associated templates use their numeric ID as the option key while choice members still use the template name.

Fixes #2839

### DIFF
--- a/test/php/library/Director/Form/IcingaTemplateChoiceFormTest.php
+++ b/test/php/library/Director/Form/IcingaTemplateChoiceFormTest.php
@@ -33,7 +33,7 @@ class IcingaTemplateChoiceFormTest extends BaseTestCase
 
         $associatedTemplates = $form
             ->getElement('required_template_id')
-            ->getAttrib('multiOptions');
+            ->getMultiOptions();
 
         $members = $form
             ->getElement('members')
--- a/test/php/library/Director/Form/IcingaTemplateChoiceFormTest.php
+++ b/test/php/library/Director/Form/IcingaTemplateChoiceFormTest.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Tests\Icinga\Module\Director\Form;
+
+use Icinga\Module\Director\Forms\IcingaTemplateChoiceForm;
+use Icinga\Module\Director\Objects\IcingaService;
+use Icinga\Module\Director\Test\BaseTestCase;
+
+class IcingaTemplateChoiceFormTest extends BaseTestCase
+{
+    protected const TEMPLATE_NAME = '___TEST___service_template_choice';
+
+    protected $templateId;
+
+    public function testServiceAssociatedTemplateUsesIdAsOptionKey()
+    {
+        if ($this->skipForMissingDb()) {
+            return;
+        }
+
+        $db = $this->getDb();
+
+        $template = IcingaService::create([
+            'object_name' => self::TEMPLATE_NAME,
+            'object_type' => 'template',
+        ], $db);
+        $template->store();
+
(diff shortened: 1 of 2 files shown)


DanielVdWP:
I have signed the CLA.

DanielVdWP:
@cla-bot check

bobapple:
@cla-bot check

DanielVdWP:
Hi @bobapple,
would it be possible to give this PR a review when you have a chance?

We now have a concrete customer use case for Service Template Choices that is currently blocked by the issue fixed here. As a temporary workaround, I am planning to provision the required choices directly in the Director database, but obviously it would be preferable to have the normal Director workflow working again.

I also think this fix could be useful beyond this specific case, since Service Template Choices are currently affected in general when an associated service template is configured.

Thanks!

raviks789 (approved):
LGTM!

raviks789 (approved):

test/php/library/Director/Form/IcingaTemplateChoiceFormTest.php
```diff
@@ -0,0 +1,72 @@
...
+        $form = IcingaTemplateChoiceForm::create('service', $db);
+        $form->setup();
+
+        $associatedTemplates = $form
+            ->getElement('required_template_id')
+            ->getMultiOptions();
+
+        $members = $form
+            ->getElement('members')
+            ->getMultiOptions();
```
raviks789: Same here.
```suggestion
            ->getAttrib('multiOptions');
```

test/php/library/Director/Form/IcingaTemplateChoiceFormTest.php
```diff
@@ -0,0 +1,72 @@
...
+        $template->store();
+
+        $this->templateId = $template->get('id');
+
+        $form = IcingaTemplateChoiceForm::create('service', $db);
+        $form->setup();
+
+        $associatedTemplates = $form
+            ->getElement('required_template_id')
+            ->getMultiOptions();
```
raviks789: PHPunit tests are failing as there is no `getMultiOptions()` method in `Zend_Form_Element` class which the  `ExtensibleSet` extends and 'multioptions' option is set through a the ExtensibleSetElement view helper via $attribs['multiOptions'] at render time. The below suggestion should resolve the failing tests.
```suggestion
            ->getAttrib('multiOptions');
```

raviks789 (changes requested):

raviks789 (approved):

Issue #3098, opened 2026-08-06
Non-existent hostgroup in role restriction breaks host creation for restricted users


0xliam:
## Describe the bug

After #2876 was resolved, users with Director hostgroup restrictions are unable to create hosts when their assigned role contains a director/filter/hostgroups restriction referencing a hostgroup that no longer exists in Director.

When creating a host from a template as a restricted user, the host creation request fails with `No such object available (UuidLookup.php:88)`.

In `library/Director/Restriction/HostgroupRestriction.php`, `listRestrictedHostgroups()` returns hostgroup names from `$this->auth->getRestrictions($this->getName())`, however this does not check if the hostgroup actually exists.

https://github.com/Icinga/icingaweb2-module-director/blob/master/library/Director/Restriction/HostgroupRestriction.php#L178-L193

## To Reproduce

1. Create a Role with a `director/filter/hostgroups` filter for an existing hostgroup - e.g.  `example-real-group` and assign it to a user.
2. Create a Role with a `director/filter/hostgroups` filter for a non-existent hostgroup - e.g.  `example-missing-group` and assign it to a user.
3. As the user with the role, Go to Director -> Hosts -> Add
4. Create a host and submit the form `example-broken-group`
5. The request fails with `No such object available (UuidLookup.php:88)` when it attempts to evaluate whether the user has access to `example-missing-group`

## Expected behavior

This may be a little too opinionated - but I would expect that Director tolerate stale `director/filter/hostgroups` restrictions: if a role references a hostgroup that has been deleted, Director should skip the missing hostgroup and continue applying restrictions based on the hostgroups that still exist. A deleted hostgroup referenced by a role should never block a restricted user from creating hosts.

## Your Environment

I am running Icinga Web on the `icinga/icingaweb2:2.14.0` container.

icingacli module list
MODULE         VERSION   STATE     DESCRIPTION
audit          1.0.3     enabled   Audit Module
businessprocess 2.6.0     enabled   A Business Process viewer and modeler
clustergraph   1.0.8     enabled   Visual representation of Icinga Clusters
director       1.11.9    enabled   Director - Config tool for Icinga 2
idoreports     0.10.1    enabled   Reports for IDO
incubator      0.23.0    enabled   Incubator provides bleeding-edge libraries
monitoring     2.12.6    enabled   Icinga monitoring module
pdfexport      0.13.0    enabled   PDF Export via Google Chrome/Chromium
reporting      1.1.0     enabled   Reporting
[...]

PR #3099, opened 2026-08-06
Skip nonexistent hostgroups in director/filter/hostgroups restrictions

0xliam:
This is my attempt at resolving #3098.

This PR adjusts `listRestrictedHostgroups()` to check each restricted hostgroup name against the `icinga_hostgroup` table, and skips names with no matching row, instead of passing every name through unchecked. This prevents a deleted hostgroup still referenced in a role restriction being evaluated by the filters.

Feedback is appreciated!

### DIFF
--- a/library/Director/Restriction/HostgroupRestriction.php
+++ b/library/Director/Restriction/HostgroupRestriction.php
@@ -60,7 +60,7 @@ class HostgroupRestriction extends ObjectRestriction
 
         // Hint: branched hosts have no id
         if (! $host->hasBeenLoadedFromDb() || $host->hasModifiedGroups() || $host->get('id') === null) {
-            foreach ($this->listRestrictedHostgroups() as $group) {
+            foreach ($this->listRestrictedHostgroupsFromDb() as $group) {
                 if ($host->hasGroup($group) || $this->matchesHostGroupFilter($group, $host)) {
                     return true;
                 }
@@ -133,14 +133,14 @@ class HostgroupRestriction extends ObjectRestriction
         IcingaObjectFilterHelper::filterByResolvedHostgroups(
             $query,
             'host',
-            $this->listRestrictedHostgroups(),
+            $this->listRestrictedHostgroupsFromDb(),
             $tableAlias
         );
     }
 
     public function filterRestrictedHostgroups(array $groups)
     {
-        return array_intersect($groups, $this->listRestrictedHostgroups());
+        return array_intersect($groups, $this->listRestrictedHostgroupsFromDb());
     }
 
     /**
@@ -158,7 +158,7 @@ class HostgroupRestriction extends ObjectRestriction
         if (! $this->isRestricted()) {
             return;
         }
-        $groups = $this->listRestrictedHostgroups();
+        $groups = $this->listRestrictedHostgroupsFromDb();
 
         if (empty($groups)) {
             $query->where('(1 = 0)');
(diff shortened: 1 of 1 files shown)


library/Director/Restriction/HostgroupRestriction.php
```diff
@@ -181,6 +181,15 @@ protected function listRestrictedHostgroups()
...
                 foreach ($this->gracefullySplitOnComma($restriction) as $group) {
+                    if (! $this->db->fetchOne(
+                        $this->db->select()
+                            ->from('icinga_hostgroup', 'id')
+                            ->where('object_name = ?', $group)
+                    ))
+                    {
+                        continue;
+                    }
+
```
raviks789: @0xliam the direction is right. But, instead of writing the query here in a for loop, I would write a separate method that filters out the restricted group against the groups in the database.

```php
    protected function listRestrictedHostgroupsFromDb()
    {
        $groups = $this->listRestrictedHostgroups();
        if (empty($groups)) {
            return [];
        }

        $query = $this->db->select()->from(
            ['icinga_hostgroup'],
            ['object_name']
        )->where('object_name IN (?)', $groups);

        return $this->db->fetchCol($query);
    }
```

And use the method `listRestrictedHostgroupsFromDb()` everywhere `listRestrictedHostgroups()` was used.

raviks789 (changes requested):

library/Director/Restriction/HostgroupRestriction.php
0xliam: Thanks for the feedback @raviks789 - I've implemented the suggested changes exactly as you wrote - I've tested this in our instance of Icinga Web and appears to be working as intended!

Appreciate any more feedback on this! :)

library/Director/Restriction/HostgroupRestriction.php
```diff
@@ -190,4 +190,28 @@ protected function listRestrictedHostgroups()
             return null;
         }
     }
+
+    /**
+     * Give a list of restricted Hostgroups that exist in the database.
+     *
+     * This filters out Hostgroups that are configured in the restriction but
+     * do not exist in Icinga DB.
```
raviks789: the hostgroups in the restriction is checked against the host groups in director database.
```suggestion
     * do not exist in Director database.
```

library/Director/Restriction/HostgroupRestriction.php
```diff
@@ -190,4 +190,28 @@ protected function listRestrictedHostgroups()
...
+     *
+     * @return array
+     */
+
+    protected function listRestrictedHostgroupsFromDb()
+    {
```
raviks789, line 203: Remove the blank line between the method and the phpdoc.
```suggestion
     * @return array
     */
    protected function listRestrictedHostgroupsFromDb()
    {
```

raviks789 (changes requested):

library/Director/Restriction/HostgroupRestriction.php
0xliam: Ah sorry, I just realised this would get confused with IcingaDB - I've pushed a fix implementing this suggested change.

library/Director/Restriction/HostgroupRestriction.php
0xliam: I've pushed a fix implementing both these suggestions!

0xliam:
Hi @raviks789  - I've implemented all suggested changes - please let me know if you'd like me to squash these commits!

raviks789 (approved):
LGTM!

raviks789:
> Hi @raviks789 - I've implemented all suggested changes - please let me know if you'd like me to squash these commits!

@0xliam you can leave the commits as they are, I can squash them while merging them into the master.

PR #3101, opened 2026-08-06
Fix pending resolve cache count

### LINKED ISSUE #3100: Housekeeping resolveCache task reports zero pending memberships
## Describe the bug

The Director housekeeping summary reports `0` pending entries for the `(Host) group resolve cache (resolveCache)` task even when the resolved host group membership cache is significantly out of sync.

In the affected environment, the membership resolver detected:

- 925 missing mappings to insert
- 1 outdated mapping to delete

However, the following command still reported a count of `0`:

```console
icingacli director housekeeping tasks

 Housekeeping task (name)                                  | Count
-----------------------------------------------------------|-------
 Undeployed configurations (oldUndeployedConfigs)          |     0
 Unused rendered files (unusedFiles)                       |     0
 Unlinked imported row sets (unlinkedImportedRowSets)      |     4
 Unlinked imported rows (unlinkedImportedRows)             |     0
 Unlinked imported properties (unlinkedImportedProperties) |     0
 (Host) group resolve cache (resolveCache)                 |     0
```

The actual resolver result was:

```json
{
    "would_insert": 925,
    "would_delete": 1
}
```

Running the native housekeeping task successfully reconciled the cache:

```console
# icingacli director housekeeping run resolveCache
```

Afterward, the resolver reported:

```json
{
    "would_insert": 0,
    "would_delete": 0
}
```

Therefore, the cache update itself works correctly. The bug affects the pending-task count shown by the housekeeping summary.

The incorrect count is caused by `Housekeeping::countResolveCache()`:

```php
public function countResolveCache()
{
    $helper = MembershipHousekeeping::instance(
        'host',
        $this->connection
    );

    return array_sum($helper->check());
}
```

`MembershipHousekeeping::check()` returns two arrays containing the new and outdated mappings. Applying `array_sum()` directly to these nested arrays does not count the contained mappings and results in an incorrect summary value.

This is operationally relevant because `getPendingTaskSummary()` and `hasPendingTasks()` rely on the task count. A stale resolved membership cache can therefore remain hidden even when hundreds of mappings are pending.

## To Reproduce

The following can be reproduced in a test environment with at least one dynamically assigned host group.

1. Configure an `IcingaHostGroup` with an `assign_filter` matching one or more hosts.
2. Create a discrepancy between the calculated memberships and `icinga_hostgroup_host_resolved`, for example by removing one resolved mapping in a disposable test database.
3. Confirm through `MembershipHousekeeping::check()` or `HostGroupMembershipResolver::checkDb()` that at least one mapping would be inserted or deleted.
4. Run:

   ```console
   icingacli director housekeeping tasks
   ```

5. Observe that `(Host) group resolve cache (resolveCache)` reports `0`.
6. Run:

   ```console
   icingacli director housekeeping run resolveCache
   ```

7. Confirm that the missing mapping is correctly restored or the outdated mapping is removed.

## Expected behavior

The `resolveCache` count should equal the total number of pending membership changes:

```php
count($newMappings) + count($outdatedMappings)
```

For example, when 925 mappings need to be inserted and 1 mapping needs to be deleted, the housekeeping summary should report:

```text
(Host) group resolve cache (resolveCache) | 926
```

The task should also be included in `getPendingTaskSummary()`, and `hasPendingTasks()` should return `true`.

## Screenshots

Not required. The relevant CLI output is included above.

## Your Environment

- Director version (System - About): `1.11.4_neteye1.40.6`
- Icinga Web 2 version and modules (System - About): `2.12.4_neteye1.149.4`; relevant modules: Director `1.11.4`, Monitoring `2.12.4`, Icinga DB Web `1.1.3`
- Icinga 2 version (`icinga2 --version`): `2.15.1`
- Operating System and version: `Red Hat Enterprise Linux 8.10 (Ootpa)`, kernel `4.18.0-553.79.1.el8_10.x86_64`
- Webserver, PHP versions: Apache HTTP Server `2.4.37`; nginx `1.24.0`; PHP CLI and PHP-FPM `7.4.33`
- Database: MariaDB
- Director objects evaluated during the full resolver test:
  - Hosts: 1820
  - Host groups: 138

## Additional context

The original symptom was an incomplete `icinga_hostgroup_host_resolved` table.

For one dynamically assigned host group:

- Icinga 2 runtime members: 71
- Director resolved-cache members: 6
- Missing Director mappings: 65

A full `resolveCache` housekeeping execution restored all 71 mappings. Across all host groups, it inserted 925 missing mappings and removed one mapping that was confirmed to be absent from the Icinga 2 runtime.

This issue is specifically about the incorrect housekeeping summary count. It is not reporting a failure of `MembershipHousekeeping::update()`, which worked correctly in the tested environment.

## Related issues

An incorrect `resolveCache` count can hide a stale resolved host group membership cache and may therefore make symptoms described in the following issues harder to diagnose:

- #1821 - `director/filter/hostgroups` ignores dynamically assigned host groups
- #2092 - hosts with applied groups are not shown in the host group member list
- #3092 - resolved host groups are not refreshed immediately after host creation or modification
- Possibly related: #2685 - inconsistent rendering of applied groups and host group members

This issue does not claim to fix the underlying behavior described in those issues. It only fixes the detection and reporting of pending `resolveCache` membership changes.

DanielVdWP:
The housekeeping summary applies `array_sum()` directly to the two arrays returned by `MembershipHousekeeping::check()`, causing pending resolve-cache membership changes to be reported incorrectly.

Count new and outdated mappings separately so that the `resolveCache` task reports the actual number of pending membership changes.

## Testing

- `git diff --check`
- PHP syntax check not run locally because PHP is not installed

Fixes #3100

### DIFF
--- a/library/Director/Db/Housekeeping.php
+++ b/library/Director/Db/Housekeeping.php
@@ -238,7 +238,9 @@ class Housekeeping
     public function countResolveCache()
     {
         $helper = MembershipHousekeeping::instance('host', $this->connection);
-        return array_sum($helper->check());
+        list($newMappings, $outdatedMappings) = $helper->check();
+
+        return count($newMappings) + count($outdatedMappings);
     }
 
     public function wipeResolveCache()
--- a/library/Director/Db/Housekeeping.php
+++ b/library/Director/Db/Housekeeping.php
@@ -238,7 +238,7 @@ class Housekeeping
     public function countResolveCache()
     {
         $helper = MembershipHousekeeping::instance('host', $this->connection);
-        list($newMappings, $outdatedMappings) = $helper->check();
+        [$newMappings, $outdatedMappings] = $helper->check();
 
         return count($newMappings) + count($outdatedMappings);
     }


library/Director/Db/Housekeeping.php
```diff
@@ -238,7 +238,9 @@ public function wipeUnlinkedImportedProperties()
     public function countResolveCache()
     {
         $helper = MembershipHousekeeping::instance('host', $this->connection);
-        return array_sum($helper->check());
+        list($newMappings, $outdatedMappings) = $helper->check();
```
raviks789:
```suggestion
        [$newMappings, $outdatedMappings] = $helper->check();
```

raviks789 (changes requested):
@DanielVdWP, the changes look good. Just a small nitpick. Use the suggested array destructuring syntax with [] instead of list().

library/Director/Db/Housekeeping.php
DanielVdWP: Updated as suggested, thanks!

raviks789 (approved):
LGTM!

PR #3109, opened 2026-09-10
Prevent infinite recursion when an Icinga Zone has a recursive parent relationship

### LINKED ISSUE #3108: Director allows the creation of infinitely recursive zones
## Describe the bug

There are two related issues I've discovered with the Director zone form and zone parent relationships:

1. The Director zone form allows a zone to select itself, or one of its descendants, as its parent, creating a recursive zone hierarchy.

For example:
```zone a -> zone-a
// or
zone-a -> zone-b -> zone-c -> zone-a
```

The form does not prevent these relationships, and there is no validation preventing a zone from becoming its own ancestor.

Once such a hierarchy exists, loading the affected zone enters infinite recursion and exhaust PHP memory:

For example:

```
object Zone "zone-a" { 
    parent = "zone-a" 
}
// or
object Zone "zone-a" {
    parent = "zone-b"
    endpoints = [ "zone-a" ]
}

object Zone "zone-b" {
    parent = "zone-c"
    endpoints = [ "zone-b" ]
}

object Zone "zone-c" {
    parent = "zone-a"
    endpoints = [ "zone-c" ]
}
```

Clicking on one of these zones results in:

```
Fatal error
: Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in
/usr/share/icinga-php/vendor/vendor/icinga/zf1/library/Zend/Db/Select.php
on line
823


Fatal error
: Uncaught ErrorException: ob_end_clean(): Failed to delete buffer. No buffer to delete in /usr/share/icingaweb2/library/Icinga/Application/Web.php:398 Stack trace: #0 [internal function]: Icinga\Application\ApplicationBootstrap->Icinga\Application\{closure}() #1 /usr/share/icingaweb2/library/Icinga/Application/Web.php(398): ob_end_clean() #2 [internal function]: Icinga\Application\Web->Icinga\Application\{closure}() #3 {main} thrown in
/usr/share/icingaweb2/library/Icinga/Application/Web.php
on line
398
```

If you try to deploy this to Icinga, it will result in:
```
critical/config: Error: Infinite recursion detected while resolving zone graph. Check your zone hierarchy.
Location: in [stage]/zones.d/zone-a/zones.conf: 1:0-1:20
[stage]/zones.d/zone-a/zones.conf(1): object Zone "zone-a" {
                                      ^^^^^^^^^^^^^^^^^^^^^
[stage]/zones.d/zone-a/zones.conf(2):     parent = "zone-a"
critical/config: Error: Infinite recursion detected while resolving zone graph. Check your zone hierarchy.
```

However, because of the infinite recursion, Director does not allow you to fix this, and the only way to resolve it is by updating the entry in the database.

## To reproduce

1. Create three zones in Icinga Director:

   * `zone-a`
   * `zone-b`
   * `zone-c`

2. Configure the zones with the following parent relationships:

   * `zone-a` → `zone-b`
   * `zone-b` → `zone-c`
   * `zone-c` → `zone-a`

3. Deploy the Director configuration.

4. Alternatively, the same condition can be reproduced with a single zone by setting its parent to itself:

   ```icinga2
   object Zone "zone-a" {
       parent = "zone-a"
   }
   ```

5. Reload or validate the resulting Icinga configuration.

6. Icinga detects the circular zone hierarchy and reports:

   ```text
   critical/config: Error: Infinite recursion detected while resolving zone graph. Check your zone hierarchy.
   Location: in [stage]/zones.d/zone-a/zones.conf: 1:0-1:20
   [stage]/zones.d/zone-a/zones.conf(1): object Zone "zone-a" {
                                         ^^^^^^^^^^^^^^^^^^^^^
   [stage]/zones.d/zone-a/zones.conf(2):     parent = "zone-a"
   critical/config: Error: Infinite recursion detected while resolving zone graph. Check your zone hierarchy.
   ```

### Expected behaviour

Icinga Director should prevent a zone from being configured with itself as its parent, or otherwise reject the configuration before generating an invalid Icinga configuration.

The same validation should prevent longer circular hierarchies such as:

```text
zone-a → zone-b → zone-c → zone-a
```

## Screenshots

Image

Image

```
MariaDB [icingaweb2_director]> SELECT id, parent_id, object_name FROM icinga_zone;
+----+-----------+-------------+
| id | parent_id | object_name |
+----+-----------+-------------+
|  5 |         5 | zone-a      |
|  6 |         5 | zone-b      |
+----+-----------+-------------+
2 rows in set (0.001 sec)
MariaDB [icingaweb2_director]> SELECT id, parent_id, object_name FROM icinga_zone;
+----+-----------+-------------+
| id | parent_id | object_name |
+----+-----------+-------------+
|  5 |         6 | zone-a      |
|  6 |         7 | zone-b      |
|  7 |         5 | zone-c      |
+----+-----------+-------------+
3 rows in set (0.001 sec)
```

## Your Environment

This has been tested on docker.io/icinga/icingaweb2:latest

Icinga Web 2 Version 	2.14.0
Git commit 	122c569fef6f01e0b7e1b9ca3c46ca2a29d63b03
PHP Version 	8.2.32
Git commit date 	2026-07-03
Loaded Libraries
icinga/icinga-php-library 	1.0.0
icinga/icinga-php-thirdparty 	1.0.0
Loaded Modules
director 		1.11.9
incubator 		0.23.0

0xliam:
Resolves #3108.

This change has three parts:

1. Prevent a zone from selecting itself as its parent in `library/Director/Objects/IcingaZone.php` by introducing a new method `Db::enumZonesExcept()`. This method is essentially the exact same as the existing `Db::enumZones` method, however it adds an `$id` parameter, which is `null` by default, and if it set, adds a `$filter` when calling `enumIcingaObjects('zone')`. 

`IcingaZoneForm` uses this method to exclude the current zone from the list of available parent zones.

Another option is to just filter out the selected zone after calling `Db::enumZones`.

2. Adds `IcingaZone::validate()` to check the complete parent hierarchy when a zone is saved. This detects both a zone selecting itself directly and indirect cycles where multiple zones eventually reference one another.

3. Change `BranchedObject::getBranchedDbObject()` to use `DbObject::getProperties()` instead of `DbObject::toPlainObject()`. This prevents the parent relationship from being resolved, which otherwise causes recursive loading when a zone hierarchy contains a cycle.

Although this fixes the memory exhaustion issue, I have not tested the change for regressions in other code paths that rely on the resolved properties returned by `DbObject::toPlainObject()`.

The recursion occurs through the following call chain:
```
IcingaObject::toPlainObject()
  -> IcingaObject::getRelatedObjectName('parent')
    -> IcingaObject::loadWithAutoIncId()
      -> DbObjectStore::load()
        -> BranchedObject::getBranchedDbObject()
          -> IcingaObject::toPlainObject()
            -> ...
```

This repeatedly resolves the `parent` relationship back to the same zone until PHP exhausts the available memory.

That last change is only required if you have an existing recursive zone hierarchy - so happy for this to be cherry picked out if you think it is not necessary.

Feedback appreciated!

### DIFF
--- a/application/forms/IcingaZoneForm.php
+++ b/application/forms/IcingaZoneForm.php
@@ -35,9 +35,10 @@ class IcingaZoneForm extends DirectorObjectForm
         $this->addElement('select', 'parent_id', array(
             'label'        => $this->translate('Parent Zone'),
             'description'  => $this->translate('Chose an (optional) parent zone'),
-            'multiOptions' => $this->optionalEnum($this->db->enumZones()),
+            'multiOptions' => $this->optionalEnum($this->db->enumZonesExcept($this->object !== null ? $this->object->get('id') : null)),
         ));
 
         $this->setButtons();
     }
 }
+
--- a/library/Director/Db.php
+++ b/library/Director/Db.php
@@ -567,6 +567,17 @@ class Db extends DbConnection
         return $this->enumIcingaObjects('zone');
     }
 
+    public function enumZonesExcept($id = null)
+    {
+        $filters = array();
+
+        if ($id !== null) {
+            $filters['id != (?)'] = (int) $id;
+        }
+
+        return $this->enumIcingaObjects('zone', $filters);
+    }
+
     public function enumNonglobalZones()
     {
         $filters = array('is_global = ?' => 'n');
--- a/library/Director/Db/Branch/BranchedObject.php
+++ b/library/Director/Db/Branch/BranchedObject.php
@@ -281,7 +281,7 @@ class BranchedObject
             }
             $branched->set('id', $this->object->get('id'));
             $branched->set('uuid', $this->object->get('uuid'));
-            foreach ((array) $this->object->toPlainObject(false, true) as $key => $value) {
+            foreach ($this->object->getProperties() as $key => $value) {
                 if ($key === 'object_type') {
                     continue;
                 }
--- a/library/Director/Objects/IcingaZone.php
+++ b/library/Director/Objects/IcingaZone.php
@@ -107,4 +107,61 @@ class IcingaZone extends IcingaObject
 
         return $this->endpointList;
     }
+    public function validate()
+    {
+        parent::validate();
+
+        $parentId = $this->get('parent_id');
+        if ($parentId === null || $parentId === '') {
+            return true;
+        }
+
+        $zoneId = $this->get('id');
+        $zoneName = $this->get('object_name');
+
+        if ($zoneId !== null && (int) $parentId === (int) $zoneId) {
+            throw new \InvalidArgumentException(sprintf(
+                'Infinite recursion detected while resolving zone graph:'
+                . ' zone "%s" cannot be its own parent. Check your zone hierarchy.',
+                $zoneName
+            ));
+        }
+
+        $path = [$zoneName];
+        $seen = [];
+        if ($zoneId !== null) {
+            $seen[(int) $zoneId] = $zoneName;
+        }
+
+        while ($parentId !== null) {
+            $parent = self::loadWithAutoIncId($parentId, $this->connection);
+            $parentName = $parent->get('object_name');
+            $path[] = $parentName;
+
+            if ($zoneId !== null && (int) $parentId === (int) $zoneId) {
+                throw new \InvalidArgumentException(sprintf(
+                    'Infinite recursion detected while resolving zone graph:'
+                    . ' setting parent to "%s" would create a cycle (%s).'
+                    . ' Check your zone hierarchy.',
+                    $path[1],
+                    implode(' -> ', $path)
+                ));
+            }
+
+            if (isset($seen[(int) $parentId])) {
+                throw new \InvalidArgumentException(sprintf(
+                    'Infinite recursion detected while resolving zone graph:'
+                    . ' pre-existing cycle in the parent chain of "%s" (%s).'
+                    . ' Check your zone hierarchy.',
+                    $zoneName,
+                    implode(' -> ', $path)
+                ));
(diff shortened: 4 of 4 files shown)


raviks789:
Closing this in favor of #3112, which fixes the same problem (#3108) the same way you diagnosed it, just with slightly different implementation choices.

For the loop check, #3112 loads all zones once and checks the whole chain in memory, instead of querying the database once per level of the hierarchy. That keeps it fast even for deep or large zone trees.

For the memory exhaustion fix, #3112 makes a smaller, more targeted change: it skips only the specific step that was resolving a zone's parent ID back into a name (which is what caused the infinite recursion), instead of switching to a different method entirely. You had actually flagged this yourself as a risk, since the broader change could affect other unrelated behavior that depends on that method.

Thanks for digging into this and finding the exact cause, it made both fixes much quicker to nail down.

0xliam:
> Closing this in favor of #3112, which fixes the same problem (#3108) the same way you diagnosed it, just with slightly different implementation choices.
> 
> For the loop check, #3112 loads all zones once and checks the whole chain in memory, instead of querying the database once per level of the hierarchy. That keeps it fast even for deep or large zone trees.
> 
> For the memory exhaustion fix, #3112 makes a smaller, more targeted change: it skips only the specific step that was resolving a zone's parent ID back into a name (which is what caused the infinite recursion), instead of switching to a different method entirely. You had actually flagged this yourself as a risk, since the broader change could affect other unrelated behavior that depends on that method.
> 
> Thanks for digging into this and finding the exact cause, it made both fixes much quicker to nail down.

Awesome, thanks for reviewing and glad that it's fixed!

PR #3111, opened 2026-09-11
CustomVariableCache: Hash checksum in PHP instead of SQL

### LINKED ISSUE #3091: MySQL SHA1 Function Issue (removed in mysql server 9.6)
## Describe the bug

Some queries in Icinga Director is using a removed SHA1() Function in MySQL 9.6 to validate some checksum. Since MySQL Version 9.6.0 the SHA1() function is disabled by default.

https://dev.mysql.com/doc/relnotes/mysql/9.7/en/news-9-6-0.html#mysqld-9-6-0-security
https://dev.mysql.com/doc/refman/9.7/en/legacy-hashing-component.html
https://dev.mysql.com/doc/refman/9.7/en/component-loading.html

I have tested it also in database:

```
mysql> SELECT SHA1('hi');
ERROR 1305 (42000): FUNCTION director.SHA1 does not exist
mysql> INSTALL COMPONENT 'file://component_classic_hashing';
mysql> SELECT SHA1('hi');
+------------------------------------------+
| SHA1('hi')                               |
+------------------------------------------+
| c22b5f9178342609428d6f51b2c5af4c0bde6a42 |
+------------------------------------------+
1 row in set (0.000 sec)
```

The matching code is here:
https://github.com/Icinga/icingaweb2-module-director/blob/master/library/Director/Db/Cache/CustomVariableCache.php#L37

When i comment this line out, it works without issues. I havn´t the depth knowledge in director and database structure to go deeper. I just want to let everybody know if they encounter this issue. 
I don´t know if it can be changed to some other more new and supported algorithm. There are serval sha1 references in the php code. 

## To Reproduce

Install a complete fresh and new instance. I use latest MySQL Server 9.7.1 in a Container and Icingaweb2 2.14.0 with Icinga Director 1.11.9 and PHP 8.4.16 on a Oracle Linux 10 (hardened) system. I use a HA Setup with two master nodes.

When i add a sample host template and a matching host, i get following error when i try to deploy the configuration based on activitiy log:

`SQLSTATE[42000]: Syntax error or access violation: 1305 FUNCTION icingadb.SHA1 does not exist, query was: SELECT v.host_id AS id, v.varname, v.varvalue, v.format, UNHEX(SHA1(v.varvalue || ';' || v.format)) AS checksum FROM icinga_host_var AS v (Pdo.php:235)`

Icinga Config via Director:
```
zones.d/master/host_templates.conf
if (! get_template(Host, "ip based device")) { template Host "ip based device" {
    check_command = "hostalive4"
    max_check_attempts = "3"
    check_interval = 1m
    retry_interval = 1m
    check_timeout = 30s
} }
object Host "testhost.local" {
    import "ip based device"

    display_name = "testhost.local"
    address = "127.0.0.1"
}
```

## Expected behavior

When Clicked on Deploy pending changes the error message come up. 

## Screenshots

Image

## Your Environment

cinga Web 2 Version	2.14.0
Git commit	122c569fef6f01e0b7e1b9ca3c46ca2a29d63b03
PHP Version	8.4.16
Git commit date	2026-07-03

Loaded Libraries
icinga/icinga-php-library	1.0.0
icinga/icinga-php-thirdparty	1.0.0
icinga/icinga-php-legacy	1.1.0

Loaded Modules
icingadb		1.4.0
businessprocess		2.6.0
cube		1.4.0
director		1.11.9
fileshipper		1.3.0
incubator		0.23.0

## Additional context

Workaround: use `INSTALL COMPONENT 'file://component_classic_hashing';` in database and it´s working fine. Maybe to add this information into documentation.

raviks789:
MySQL 9.6 disabled SHA1() by default, which broke deploys with "FUNCTION ... SHA1 does not exist". This checksum was only ever used as an in-memory cache key while rendering config, never stored, so there was no reason to compute it in the database at all. Hashing it in PHP after the row is fetched drops that dependency on MySQL's SHA1() and the pgcrypto extension on Postgres.

fixes #3091


lippserd:
Good idea. But we could adjust docs right away and remove `pgcrypto` everywhere then right?

raviks789:
> Good idea. But we could adjust docs right away and remove `pgcrypto` everywhere then right?

You're right. I will make the necessary changes accordingly.

lippserd (approved):

PR #3113, opened 2026-09-14
Fix command inheritance not shown

### LINKED ISSUE #3007: Imported command is marked “Not in use” and variables are not shown in Overview

raviks789:
fixes #3007


library/Director/Resolver/CommandUsage.php
```diff
@@ -60,6 +61,10 @@ public function getLinks()
                 'template' => $this->translate('%d Notification Template(s)'),
                 'apply'     => $this->translate('%d Notification Apply Rule(s)'),
             ],
+            'command' => [
+                'object'   => $this->translate('%d Commands(s)'),
+                'template'   => $this->translate('%d Command Templates(s)'),
```
lippserd:
```suggestion
                'object'   => $this->translate('%d Command(s)'),
                'template' => $this->translate('%d Command Template(s)'),
```

lippserd (changes requested):
Counting imports makes used command templates fail to open in the editor. The new inheritance count in `IcingaCommand::countDirectUses()` makes `isInUse()` true for templates, so `CommandController::showUsage()` constructs `CommandUsage`, whose constructor rejects templates.

I reproduced this in the browser: create a command template, import it into a command, then open the template. The form is replaced by `CommandUsageTable expects object or external_object, got a template`. Removing the importing command restores the page.

Please make the controller and resolver handle used templates consistently, either by supporting templates in the resolver or by using the existing template usage mechanism. Add a regression check that opens an imported template and verifies its form and usage.

lippserd (approved):

PR #3114, opened 2026-09-15
Split event command into its own section, show it on objects too

### LINKED ISSUE #2531: Event Command not shown when editing a service
## Expected Behavior
When a new service is created and at least one "event command" is defined, it can be selected under "Check execution" (I'm wondering if this is the correct title?) in the field "Event command".
After storing the new service I'd expect that the chosen event command can be changed and the service could be stored again.

## Current Behavior
Currently the defined event command is not displayed anymore after the service has been stored the first time. It is only shown in the "Preview"-tab of the service. This means it cannot be modified anymore. Also cloning the service doesn't help.

1. define a event command: Director -> Commands -> Commands -> Add -> Command type: “Event Plugin Command”
2. define a single service: Director -> Services -> Add, expand "Check execution" and klick in  the empty "Event command"-field to choose the defined new event command. Klick "Store"
3. klick on the new service: you only see "Main properties", "Custom properties", "Additional properties" - "Check execution" is missing. Klick on the "Preview"-tab of the service: you'll see "event_command = "[your-command-name]"

## Your Environment

* Director version (System - About): 1.9.1
* Icinga Web 2 version and modules (System - About): 2.8.2
* Icinga 2 version (`icinga2 --version`):  r2.13.1-1
* Operating System and version: SLES 15 SP3
* Webserver, PHP versions: Apache/2.4.51, PHP 7.2.5

raviks789:
Event command used to sit inside the check execution section and only showed up when editing a host or service template. Now it has its own section, and shows up on regular hosts and services as well, not just their templates.

resolves #2531


library/Director/Web/Form/DirectorObjectForm.php
```diff
@@ -1394,18 +1409,37 @@ protected function addCheckCommandElements($force = false)
...
-            $this->addElement('select', 'event_command_id', array(
-                'label' => $this->translate('Event command'),
-                'description'  => $this->translate('Event command definition'),
-                'multiOptions' => $this->optionalEnum($eventCommands),
-                'class'        => 'autosubmit',
-            ));
-            $this->addToCheckExecutionDisplayGroup('event_command_id');
+        if (empty($eventCommands)) {
+            return $this;
         }
```
lippserd: Keep `enable_event_handler` available when no event commands exist. This return now hides the enable/disable control as well as the command selector, whereas templates previously exposed the boolean independently. I reproduced this on host and service templates with a stored `false` value: the setting disappears from the form and cannot be changed or cleared. Only conditionally add `event_command_id`, and always add and group `enable_event_handler`.

lippserd (changes requested):

library/Director/Web/Form/DirectorObjectForm.php
```diff
@@ -24,6 +24,7 @@
 use Icinga\Module\Director\Web\Form\Validate\NamePattern;
 use Zend_Form_Element as ZfElement;
 use Zend_Form_Element_Select as ZfSelect;
+use Zend_Form_Exception;
```
lippserd, line 28:
```suggestion
use Zend_Form_DisplayGroup;
use Zend_Form_Exception;
```

library/Director/Web/Form/DirectorObjectForm.php
```diff
@@ -340,6 +342,16 @@ protected function addToCheckExecutionDisplayGroup($elements)
         );
     }
 
+    protected function addToEventHandlerDisplayGroup($elements)
```
lippserd:
```suggestion
    /**
     * Add named elements to the event handler display group
     *
     * @param string ...$elements Names of existing form elements
     *
     * @return Zend_Form_DisplayGroup
     *
     * @throws Zend_Form_Exception If a new group has no valid elements
     */
    protected function addToEventHandlerDisplayGroup(string ...$elements): Zend_Form_DisplayGroup
```

library/Director/Web/Form/DirectorObjectForm.php
```diff
@@ -1394,18 +1409,40 @@ protected function addCheckCommandElements($force = false)
...
+            $el->setAttrib('title', $this->translate('There are no event commands available'));
         }
 
+        $this->optionalBoolean(
+            'enable_event_handler',
+            $this->translate('Enable event handler'),
+            $this->translate('Whether to enable event handlers this object')
+        );
+
+        $this->addToEventHandlerDisplayGroup(['event_command_id', 'enable_event_handler']);
```
lippserd:
```suggestion
        $this->addToEventHandlerDisplayGroup('event_command_id', 'enable_event_handler');
```

library/Director/Web/Form/DirectorObjectForm.php
```diff
@@ -1394,18 +1409,40 @@ protected function addCheckCommandElements($force = false)
...
 
+        return $this;
+    }
+
+    /**
+     * Add the event command fields
+     *
+     * @return $this
+     */
+    protected function addEventCommandElements()
```
lippserd:
```suggestion
    protected function addEventCommandElements(): static
```

lippserd (approved):

lippserd (approved):

